PR #188, opened 2013-03-30
Added email and github to python.snippte-docs

jbcurtin:
Just added email and github as snips_email and snips_github


laxtiz:
Why not define default values
I hate this.
anytime i want entend docs give me a error.
 then just know you must have g:snips_*

lpil:
I agree with laxtiz, snippets should all work out-of-the-box.

PR #501, opened 2015-01-05
Laravel Snippet

FiloSpaTeam:
Snippet useful for Laravel Framework (Php).
Contains:
* Resource Controller
* Service Provider for Service
* Service Provider for Repository
* Base Model
* Base Repository
* Base Service
* Facade


SirVer:
Does laravel have it's own `filetype` we could use? Most php users will probably not be enthused about having more snippets that pollute the namespace.

FiloSpaTeam:
No, vim return php filetype :(
I can do this:
* Create a function that set correct filetype from a user input when triggered FileType event
* Add a function that read into the project tree or something like that, that search for a file called (project.filetype for example) that contains the correct file type for the original filetype (an example: php laravel.php)
* Idk... XD
Otherwise laravel snippets need php_laravel notation...

UltiSnips/php_laravel.snippets
```diff
@@ -0,0 +1,260 @@
+#resource controller
+snippet l_rsc "Laravel resource controller"
```
SirVer: most (all?) of the snippets should have the "b" option set.

FiloSpaTeam: Ok now correct all snippets :+1: 
I've not seen this option xD sorry.

SirVer:
Thanks.

PR #502, opened 2015-01-05
Angular snippets updated to build your param list automatically

nerdwaller:
The old way requires you to type your parameter annotations and manually list your parameters, this new method will automatically populate valid parameters from your annotated parameter list.  The assumption I am making is that most use cases use the same names as parameters (seems safe from most idiomatic angular I have seen).

Screen gif below (I use a different binding locally, but left the original one here so users aren't thrown in the cold):

angularexpand


SirVer:
Great changes! Thanks.

PR #503, opened 2015-01-06
Add iem python snippet for embedded ipython

yikelu:
I use this quite often to embed ipython interpreter almost like a break point.


honza:
Thanks! :sparkles:

PR #504, opened 2015-01-07
Snippet for python prompt toolkit embed

tony:
https://github.com/jonathanslenders/python-prompt-toolkit

```python
from prompt_toolkit.contrib.repl import embed
embed(globals(), locals(), vi_mode=False, history_filename=None)
```


SirVer:
thanks.

PR #506, opened 2015-01-17
add Ada GMGPL snippet; add AWK snippets

m-pilia:


lpil:
Why are there licence snippets in the Ada snippets file?

And wouldn't most of these be better in the snipmate format? They are generally not using ultisnips features.

m-pilia:
Hi. For license snippets, I thought it is normal, I see the GPL in c.snippets too. I searched but I did not found indications about such kind of snippet organization. Should I move license snippets into some other file?

About the snipmate vs. UltiSnips question, I wrote the same snippets in both formats. I read some previous discussions (but I'm still quite confused on this matter) and I concluded both format should be supported on this repo. Should I write easy snippets in snipmate only, and complex snippets (needing Python support) in UltiSnips? It's quite hard to deduce, looking at current code in the repo, previous discussions and README guidelines.

There is no problem for me to rewrite my commits. But maybe you should clarify guidelines on such matters in the README file, it would simplify a lot the work for occasional contributors.

lpil:
Well we have licence snippets in `_.snippets`, which should be accessable from all filetypes. Perhaps see if these are sufficient?

That is my approach, yes. Ultisnips loads both types, and favours the Ultisnips format ones when there is a name collision. If they are functionally the same, you may as well just write the snipmate format :)

A good idea. I'll have a look at the README. Thanks!

PR #508, opened 2015-01-18
[puppet.snippet]: s/'/"/ in most of snippets, changes in 'file' snippet ...

slebedev:
...and some new snippets

### DIFF
--- a/snippets/puppet.snippets
+++ b/snippets/puppet.snippets
@@ -90,7 +90,7 @@ snippet class
 		${0}
 	}
 snippet node
-	node '${1:`vim_snippets#Filename('', 'fqdn')`}' {
+	node "${1:`vim_snippets#Filename('', 'fqdn')`}" {
 		${0}
 	}
 snippet case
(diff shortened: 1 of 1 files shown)


lpil:
Why are you swapping `"` for `'`?  It is a popular ruby convention to only use `"` when you require the extra formatting features that `"` offers. We aim to stick to conventions where possible.

slebedev:
Variables wouldn't resolve between `'`. For example:

```
$root_dir = '/opt/myApp'

file { '$root_dir' :
...
```

Resource file in this case will be `$root_dir` instead of `/opt/myApp`. To avoid this you must either delete `'` after snippet resolve (`file { $root_dir :`), or replace `'` to `"` in snippet. I think second way is more correct, it'll be able to use variables in string (for ex.: `file { "$root_dir/$module" :`)

lpil:
That's not Ruby string interpolation, was I wrong in thinking that puppet is a Ruby DSL?

slebedev:
Generally - yes, but see https://docs.puppetlabs.com/learning/manifests.html#syntax-hints

slebedev:
Quote from page above:

>There are two kinds of quotes in Puppet: single (') and double ("). The main difference is that double quotes let you interpolate $variables, which we cover in another lesson.

slebedev:
And https://docs.puppetlabs.com/puppet/latest/reference/lang_datatypes.html#strings

slebedev:
Hello? Are there somebody? :(

SirVer:
@lpil do you have any more concerns about this? To me it looks good, but I do not speak ruby and never used puppet.

lpil:
I think I was wrong in thinking that it's a ruby DSL, rather it seems to be a ruby-like language implemented in ruby. As a result I don't think that the ruby community style guidelines hold here. It'd be good to have someone who knows puppet weigh in, but otherwise I don't see any problems here.

PR #510, opened 2015-01-24
erlang: remove spaces before parens in UltiSnips

ches:
So this is ultimately a matter of style without a strict standard, but I have pretty much never seen anyone write Erlang in the style previously used by the UltiSnips snippets (a space preceding the opening parenthesis of directives, and even the parameters of function definitions)—the official Erlang documentation doesn't use this style, code skeletons in [the Emacs mode shipped in the Erlang/OTP distribution][1] (also borrowed by vimerl) don't, and if it's a clincher, the Snipmate snippets in this repository don't either.

I've had local overrides for awhile, and finally decided everyone else shouldn't have to do the same.

Also adds a list comprehension snippet, as a bonus :smile: 

**An aside question** for @SirVer while I'm here, if I may: see the `fu` Erlang snippet here as an example—the third placeholder (with nested fourth) is meant to be an optional guard clause that you might backspace over if it is unneeded. However, if I backspace over the *second* placeholder in order to define a function with no arguments, moving to the third placeholder shifts the text selection to the wrong place. I see this behavior regularly with other UltiSnips snippets. Perhaps it's something with my Vim config that I should try a minimal configuration to reproduce, but is this a familiar issue, and is there any Vim config or snippet definition practice that avoids it? Thanks!

[1]: http://www.erlang.org/doc/apps/tools/erlang_mode_chapter.html

### DIFF
--- a/UltiSnips/erlang.snippets
+++ b/UltiSnips/erlang.snippets
@@ -5,12 +5,12 @@
 priority -50
 
 snippet pat "Case:Receive:Try Clause"
-${1:pattern}${2: when ${3:guard}} ->;
+${1:pattern}${2: when ${3:guard}} ->
 	${4:body}
 endsnippet
 
-snippet beh "Behaviour Directive"
--behaviour (${1:behaviour}).
+snippet beh "Behaviour Directive" b
+-behaviour(${1:behaviour}).
 endsnippet
 
 snippet case "Case Expression"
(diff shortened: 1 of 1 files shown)


SirVer:
Changes lgtm. Thanks.

What you describe sounds like a bug. There are tests for this that pass on my system, so I assume you have some corner case in your setup or some bad interplay with another plugin. A minimal repro case on a bug report would be appreciated.

PR #511, opened 2015-01-25
Fix false positive hyphen list item expansion.

### LINKED ISSUE #491: Markdown three dashes `---` ATX header conflicts with list syntax
Initial state:

```
-<cursor>
```

When I do `<tab>`, I expect it to become:

```
-   <cursor>
```

(3 spaces) which is the original and most portable way to write lists (although many modern implementations work fine with a single space).

But instead it interprets that as a prefix of `---` https://github.com/honza/vim-snippets/blob/d00ca9f573405e61ea4af5a2101e7b697d6bd8be/snippets/markdown.snippets#L74 and tries to expand that.

Is there a way to require two or more dashes `--` for the header expansion?

cirosantilli:
Fix #491.

### DIFF
--- a/snippets/markdown.snippets
+++ b/snippets/markdown.snippets
@@ -71,6 +71,8 @@ snippet ===
 	`repeat('=', strlen(getline(line(".") - 1)) - strlen(getline('.')))`
 
 	${0}
+snippet -
+	-   ${0}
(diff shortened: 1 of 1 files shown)


SirVer:
Thanks.

PR #512, opened 2015-01-26
Added snippet for remote python debugger (rpdb)

AndreaGiardini:


lpil:
Hello! Thanks for this.

Can you remove the Ultisnips version? Ultisnips will also read the snipmate format, and we would like to avoid unneeded duplication.

AndreaGiardini:
Should be ok now :)

lpil:
Thanks! :)

PR #514, opened 2015-01-30
Update PHP getter setter

jdeniau:
Add return / parameter types
Set @param and @return


SirVer:
Nice change!

jdeniau:
Thank you !

PR #515, opened 2015-01-31
Add a missing closing double quote in HTML

tszym:
Hi, I just found a missing closing double-quote in this HTML snippet for an input.


lpil:
Thanks! :)

PR #516, opened 2015-02-03
Update PHP class and interface snippets

jdeniau:
Include <?php and namespace automatically
Add namespace snippet

### DIFF
--- a/UltiSnips/php.snippets
+++ b/UltiSnips/php.snippets
@@ -261,35 +261,55 @@ if (${1:/* condition */}) {
 $0
 endsnippet
 
+snippet ns "namespace declaration" b
+namespace ${1:`!p
+abspath = os.path.abspath(path)
+m = re.search(r'[A-Z].+(?=/)', abspath)
+if m:
+›   snip.rv = m.group().replace('/', '\\')
+`};
+endsnippet
+
 snippet class "Class declaration template" b
+<?php
+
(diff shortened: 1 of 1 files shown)


lpil:
Thanks :)

WouterSioen:
It seems like this adds the full path to the class as namespace. Is there a possibility to only include the relevant path from the project root? Or is this too complicated for snippets?

lpil:
It's python so you can do whatever you want :)

WouterSioen:
Ok, let's polish my python skills. Thanks!

jdeniau:
The problem here is to define the path referencing your current path.

The best things to do is maybe to work with composer here, but I don't know if it's possible and easy to implement.

codeinabox:
Did anything happen with the way name spaces are generated? Currently I get things like this

```php
namespace Users\matt\workspace\someproject\src\MyBundle\Document;
```

When I'd rather it generated the following

```php
namespace MyBundle\Document;
```

Happy to help out with a PR

lpil:
PRs very welcome!

codeinabox:
I think this will do the trick #720

codeinabox:
I also updated the Laravel, Symfony and PHPUnit snippets #721

WouterSioen:
I just tested this and it works really great. Big thanks @codeinabox!

PR #518, opened 2015-02-10
fix :f snippet skipping over function body

leonidez:


lpil:
Hi Leonidez. Thanks for this. I think the existing behaviour with the comma is a bit strange.

Could you drop the `:` from 3 please? :)

leonidez:
Perhaps a literal comma at the end would be better? Since we are creating a method that would be the most common use case.

lpil:
What do you mean? :)

leonidez:
You mentioned the strange behavior with the comma. I assumed you meant that ',' is added as a hint at the end of the snippet. Is there some other behavior you're referring to?

lpil:
Yes, that is what I mean. I'm not really a fan of such small hints as it can be hard to tell if you're going to overwrite it or not when you type. It's especially hard with a single character.

I would prefer either there to be no comma, or for the comma to be before the closing `$0`

What are your thoughts?

leonidez:
I like having the comma there. I almost always need one in this context.

Also, most of the snippets do not have a tab after the closing of a function block. Perhaps it should simply be a comma literal and the function body has the last tab. I could pull request that if it's favorable.

lpil:
Sounds good to me! Push another commit making this change please :)

leonidez:
Done. Thanks man.

lpil:
Great, thanks :)

PR #520, opened 2015-02-13
Addedd incl snippet for cpp

tapanprakasht:
I have created  incl snippet for cpp which enables to include header files easily.

### DIFF
--- a/snippets/cpp.snippets
+++ b/snippets/cpp.snippets
@@ -129,3 +129,7 @@ snippet lld
 	[${1}](${2}){
 		${3}
 	}
+snippet incl
+	#include<iostream>
+	using namespace std;
+	${1}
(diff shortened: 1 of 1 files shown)


lpil:
Is this general enough? Wouldn't people want to incl other things and use other name spaces?

tapanprakasht:
Added uname snippet for adding namespaces.

snippets/cpp.snippets
```diff
@@ -129,3 +129,6 @@ snippet lld
 	[${1}](${2}){
 		${3}
 	}
+snippet uname
```
SirVer: Using a namespace is relatively short, I do not think it needs a snippet. Making one is here
See https://github.com/honza/vim-snippets/blob/master/UltiSnips/cpp.snippets#L27.

Should use the same trigger, the UltiSnips one is much superior though.

SirVer:
ping?

SirVer:
closing for no reply.

PR #521, opened 2015-02-18
just added a 'use' snippet for elixir

craigp:


lpil:
This snippet doesn't actually save any keypresses than just typing it. In fact it uses more with the default Ultisnips settings.

Why do you want this?

craigp:
Yeah, you're right, just can it, was't thinking

lpil:
:) Thanks for the pull request anyway. Do send us another if you think of anything else.

PR #523, opened 2015-02-24
Fixed some PHP Code Standards

mjacobus:
I just improved the "new", "def" snippets and fixed some spaces to be PSR-2 compliant.

### DIFF
--- a/UltiSnips/php.snippets
+++ b/UltiSnips/php.snippets
@@ -8,7 +8,7 @@ $${1:arrayName} = array('${2}' => ${3});${4}
 endsnippet
 
 snippet def "def"
-define('${1:FOO}', ${2:'bar'});${3}
+define('${1:VARIABLE_NAME}', ${2:'definition'});${3}
 endsnippet
 
 snippet do "do"
@@ -253,7 +253,7 @@ $0
 endsnippet
 
 snippet new "New class instance" b
-$${1:foo} = new ${2:${1/\w+\s*/\u$0/}}($3);
+$${1:variableName} = new ${2:${1/\w+\s*/\u$0/}}($3);
 $0
 endsnippet
 
--- a/UltiSnips/php.snippets
+++ b/UltiSnips/php.snippets
@@ -8,7 +8,7 @@ $${1:arrayName} = array('${2}' => ${3});${4}
 endsnippet
 
 snippet def "def"
-define('${1}'${2});${3}
+define('${1:FOO}', ${2:'bar'});${3}
 endsnippet
 
 snippet do "do"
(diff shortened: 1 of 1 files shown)


UltiSnips/php.snippets
```diff
@@ -8,7 +8,7 @@ $${1:arrayName} = array('${2}' => ${3});${4}
 endsnippet
 
 snippet def "def"
-define('${1}'${2});${3}
+define('${1:FOO}', ${2:'bar'});${3}
```
lpil: Please don't use non-documentary placeholder text.

UltiSnips/php.snippets
```diff
@@ -240,14 +246,14 @@ $0
...
-foreach ($${1:variable} as $${3:value}){
+foreach ($${1:variable} as $${3:value}) {
 	${VISUAL}${4}
 }
 $0
 endsnippet
 
 snippet new "New class instance" b
-$$1 = new $1($2);
+$${1:foo} = new ${2:${1/\w+\s*/\u$0/}}($3);
```
lpil: Here too!

lpil:
See my notes above :)

SirVer:
@lpil Louis, I just learned that snipMate now has a "version 2" file format. I think we should reconsider what lives in `snippets/`, because neither NeoSnippet nor UltiSnips will correctly deal with version 2 snippets. Maybe it is time to split the repo again.

UltiSnips/php.snippets
mjacobus: I'm sorry, what is a "non-documentary placeholder text"?

SirVer: FOO, TODO, BLAH, BLUB, /* code */. Just make `${1}`.

lpil:
That's annoying- we're going to have a lot of duplication. Can we have another folder for snipmate v2 snippets?

One for snipmate format, one for Ultisnips format? What format does NeoSnippet use?

Do these repos already exist? How do we communicate this change to users? Seems like it could be a bit of a painful transition.

UltiSnips/php.snippets
mjacobus: All right. Fixed. I thought FOOs and BARs were pretty straightforward though. =P

SirVer:
> That's annoying- we're going to have a lot of duplication. Can we have another folder for snipmate v2 snippets?

dunno. I think snipMate did not change their search path. @garbas, can you comment? 

>  What format does NeoSnippet use?

I think it only parses v1 snipMate snippets. @Shougo, can you comment?

ajzafar:
As I stated in #528, this repo need change nothing until contributors decide otherwise.

ajzafar:
Or add `version 0` to the top of SnipMate snippets files.

SirVer:
Reviewed files:
- UltiSnips/php.snippets @ r2

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/honza/vim-snippets/523)

SirVer:
All good. Gonna merge. Thanks!

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/honza/vim-snippets/523)

PR #525, opened 2015-02-24
JS snippet: anonymous function variable assignment.

### LINKED ISSUE #524: JavaScript Snippet for var f = function () {}
I see a snippet for assigning a named function to a local variable, but none for the (more popular in my mind) anonymous function assignment to a local var. Is this intentionally omitted, or just in need of a PR? =)

I was thinking `vaf` might be appropriate (var anonymous function)?

nybblr:
Closes #524

@lpil anything else?

### DIFF
--- a/snippets/javascript/javascript.snippets
+++ b/snippets/javascript/javascript.snippets
@@ -15,6 +15,11 @@ snippet f
 	function(${1}) {
 		${0}
 	}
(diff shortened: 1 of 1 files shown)


lpil:
Great! Thanks nybblr. If you find anything else we can improve with the JS snippets, do let us know!

PR #527, opened 2015-02-28
changed snippet getopts to work correctly

switch87:
the original snippet gave some errors, I changed it a bit to have it working


honza:
Thanks! :sparkles:

PR #529, opened 2015-03-01
repeared shebang

switch87:
changed shebang snippet, did not work (becous of the leading # in the snippet?)
also changed the function snippet becous it was verry confusing, sh version could be placed back, but snippet for bash was confusing without looking at the plugin-file.

### DIFF
--- a/snippets/sh.snippets
+++ b/snippets/sh.snippets
@@ -1,6 +1,6 @@
 # Shebang. Executing bash via /usr/bin/env makes scripts more portable.
 snippet bash
-	#!/usr/bin/bash
+	#!/bin/bash
 
 snippet if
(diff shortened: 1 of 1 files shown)


honza:
Thanks :sparkles:

lpil:
Why have you removed the env shebang? This is more portable and flexible than using bash directly.

The comment is also wrong now.

switch87:
#!/bin/bash is not wrong, its the most used version.
the change of function you dit is wrong if u use bash for shebang completion, this would be for shm what is not included in the snippets.

lpil:
It is not wrong, but it is much more concrete than env. I would recommend against using it, but if you want a snippet for that anyway do not remove the env snippet.

PR #530, opened 2015-03-02
Revert "repeared shebang"

lpil:
Reverts honza/vim-snippets#529


honza:
Did I jump the gun?

lpil:
Maybe a little. The commit removed a good snippet without reason, and left a comment incorrect.

PR #531, opened 2015-03-02
same changes but with remark of lpil in thougt

switch87:


lpil:
Quick question- why are `sh` snippets using the `bash` runtime, rather than `sh`? Perhaps the `#!` snippet should use `sh`, and the bash one should use `bash`

Your branch needs rebasing before I can merge it! :)

switch87:
Sorry, I am quite new to git(hub), will do the rebasing. I will add the sh as well, I was asking the same question, but left it as is becouse I only use bash for the moment.

switch87:
I made a new fork becous I could not resolve the problem with git, so I will close this pull request.

lpil:
OK :)

PR #532, opened 2015-03-02
changed shebangs end function snippet

switch87:
Same pull request as before, but could not resolve rebasing problem.


lpil:
Looks good :)

PR #534, opened 2015-03-04
Created new ui5 framework snippets

ricardogcolombo:
i've created a new file for ui5 snippets with some of all the clases in the framework.


lpil:
Hi Richardo!

I'm reluctant to add these framework specific snippets with this current filename as it'll result in the snippet being loaded for everyone, even those not using the framework. The Javascript snippets namespace is already very busy!
This is especially an issue since the framework in question is not very widely used. Might I suggest renaming to `openui5.js.snippets`, and then adding something to your vimrc to instruct your plugin to load them for you?

(also, you don't seem to have correctly set up your email on your github account, it's not recognising the one on your commits, and thus is not attributing the commits to your account)

Cheers,
Louis

SirVer:
@Ipil @ricardogcolombo Guys, with the new vendor approach in place, does it make sense to revisit this PR?

ricardogcolombo:
@SirVer i think that the framework is growing and is used by so many companies in their sites, most of that uses SAP, would be great give to other developers an option to the Java IDEs. 

For my side im open to continue adding different clases on the snippet file.

lpil:
Sounds fine as long as we move it out of the default namespace.

SirVer:
@ricardogcolombo I think this will make a contribution, but the snippets should be hoisted out of the default namespace for everybody. Please see https://github.com/honza/vim-snippets#vendor-snippets on how to do that.

ricardogcolombo:
Hi Guys, sorry the delay but i was offline for a long time. I've moved the file to the UltiSnips folder and renamed as javascript_openui5.snippets. Please let me know if something is missing, i will update the file ASAP with more snippets. 

Thank you.

SirVer:
Not sure about the indentation in these files.... But I do not use the library, so I cannot really comment. Please rename the file to `javascript-openui5.snippets` so that it does not get pulled in for each javascript file.

SirVer:
@ricardogcolombo Could you do the rename? The rest lgtm.

SirVer:
Thanks for your work on that, Ricardo!

PR #535, opened 2015-03-06
Add h2 to h6 tags

filipegiusti:
Almost a straight copy/paste from `h1`.


lpil:
These snippets are already in the snippet directory, and don't need to be duplicated in the ultisnips directory. THanks though!

filipegiusti:
See https://github.com/honza/vim-snippets/pull/536

PR #536, opened 2015-03-06
Add HTML5 tags support

filipegiusti:
I copied all html5 tags from `snippets/html.snippets`, seems to be complete.

### DIFF
--- a/UltiSnips/html.snippets
+++ b/UltiSnips/html.snippets
@@ -291,4 +291,225 @@ snippet movie "Embed QT movie (movie)" b
 	`!p x(snip)`>
 </object>
 endsnippet
+
+##############
+# HTML5 TAGS #
+##############
+
+snippet article "article" w
+<article>
+	$0
+</article>
+endsnippet
+
+snippet article. "" w
+<article class="$1">
+	$0
+</article>
+endsnippet
+
+snippet article# "" w
+<article id="$1">
+	$0
+</article>
+endsnippet
+
+snippet aside "" w
+<aside>
+	$0
+</aside>
+endsnippet
+
+snippet aside. "" w
+<aside class="$1">
+	$0
+</aside>
+endsnippet
+
+snippet aside# "" w
+<aside id="$1">
+	$0
+</aside>
+endsnippet
+
+snippet audio "" w
+<audio src="$1>$0</audio>
+endsnippet
+
+snippet bdi "" w
+<bdi>$0</bdo>
+endsnippet
+
+snippet canvas "" w
+<canvas>
+	$0
+</canvas>
+endsnippet
+
+snippet details "" w
+<details>$0</details>
+endsnippet
+
+snippet dialog "" w
+<dialog>
+	$0
+</dialog>
+endsnippet
+
+snippet embed "" w
+<embed src="$1" type="$0" />
+endsnippet
+
+snippet figcaption "" w
+<figcaption>$0</figcaption>
+endsnippet
+
+snippet figure "" w
+<figure>$0</figure>
+endsnippet
+
+snippet figure# "" w
+<figure id="$1">
+	$0
+</figure>
+endsnippet
+
+snippet figure. "" w
+<figure class="$1">
+	$0
+</figure>
+endsnippet
+
+snippet footer "" w
+<footer>
+	$0
+</footer>
+endsnippet
+
+snippet footer. "" w
+<footer class="$1">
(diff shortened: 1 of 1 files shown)


lpil:
Ultisnips should read from the snippets directory already, so there is no need to do this. Did you have problems with these snippets not being available?

filipegiusti:
Yes. I didn't know that it should read from both.

I was editing an `eruby` file when I noticed the problem. I tried it now and only `UltiSnips/eruby.snippets`, `UltiSnips/html.snippets` and `snippets/eruby.snippets` are available but not `snippets/html.snippets`. However if I do start an html file, both `UltiSnips/html.snippets` and `snippets/html.snippets` are available. Have you ever experienced something like that? Can you point any direction where I should look to fix it?

Also when having duplicates it's preferred to keep them on UltiSnips or snippets?

lpil:
The preference is to keep everything in snippets, as this is the most widely read format. This might change now though, since there is a new version...

@SirVer is the man to talk to about Ultisnips behaviour. :)

filipegiusti:
@SirVer is that a problem on UltiSnips? Should I open an issue?

SirVer:
@filipegiusti Not a problem per-se. snipMate does not support the 'extends' keyword, so UltiSnips does not emulate it for snipMate. The argument could be made that extends should be snippet source agnostic, but that will mean that snipMate users will see other snippets than UlitSnips users for the same snippet/ directory.

What I mean is that US erred on the side of better compatibility with snipMate. SnipMate however has developed further and now has no longer a clear feature set - there are fuzzy borders between version 1 and 2 and I am unsure what UltiSnips should emulate or, if it even should emulate snipMate at all still.

filipegiusti:
@SirVer it appears that snipmate supports `extends` since [0.87 - 2014-01-04](https://github.com/garbas/vim-snipmate#087---2014-01-04) it also has the alternative of using `scope_aliases` in the [FAQ](https://github.com/garbas/vim-snipmate#faq). However as you said, there is no clear feature set.

Right now I'm keeping a local copy with the snippets I need, but it would be cool to work something out with both UltiSnips and vim-snippets. Either vim-snippets choose to keep UltiSnips snippets as the main ones, or UltiSnips implement one way to express dependencies.

Let me know your thoughts @SirVer and @lpil .

ajzafar:
@SirVer I'm aiming for SnipMate's help file to be the definitive and complete specification of what SnipMate does/supports. If you have any further questions, you're welcome to contact me, and I'll try to clarify the help file. It's not perfect, I know.

And yes, SnipMate currently supports the `extends` keyword.

SirVer:
Thanks for your comment. I am actually surprised that new features have been added to snipMate at all - my understanding was that it was an abandoned plugin and UltiSnips the transition path fully replacing it. That was the reason I added snipMate support - to make it easier for people to move to UltiSnips. I also thought there was consensus on this decision, but apparently I was wrong.

I am not sure what is advisable for UltiSnips to do at this point and I am very interested in opinions:

1)  add support for version [0|1] snipMate files, extends in snipMate and conditionally supporting nested tabstops depending on version. That means that UltiSnips would need to keep track of diverting implementations in the future too.
2) keep partial snipMate support as is, maybe add extends only
3) remove snipMate support, demerge the snippets again and develop itself further.

What would you advise? Are there other solutions?

ajzafar:
> Thanks for your comment. I am actually surprised that new features have been added to snipMate at all - my understanding was that it was an abandoned plugin and UltiSnips the transition path fully replacing it. That was the reason I added snipMate support - to make it easier for people to move to UltiSnips. I also thought there was consensus on this decision, but apparently I was wrong.

I'm very sorry if I gave that impression. I'm sure the pace of development hasn't helped either. I intend to improve SnipMate as best I can while maintaining as much simplicity as possible.

> I am not sure what is advisable for UltiSnips to do at this point and I am very interested in opinions:
> ...
> What would you advise? Are there other solutions?

I guess I don't have much of an opinion here. If it helps though, going forwards there shouldn't be any changes at all to version 0. Additionally, should anyone have any suggestions regarding changes to SnipMate that'll help in compatibility, I am open to them.

SirVer:
> I guess I don't have much of an opinion here. If it helps though, going forwards there shouldn't be any changes at all to version 0.

seems like keeping all snipMate snippets in vim-snippets at version 0 then should help. Then only the extends keyword need to be sorted out.

PR #537, opened 2015-03-06
Correct img tag

czzshr:
I delete the blank before the "img" tag, because it's redundant when Vim could indent automatically.


lpil:
Thanks!

PR #538, opened 2015-03-08
Add main element

mrzool:
From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main):

> The HTML `<main>` element represents the main content of  the `<body>` of a document or application.
> ...
> The `<main>` element is widely supported (except for Internet Explorer). It is suggested that until `<main>` element is supported in Internet Explorer, the "main" ARIA role be added to the `<main>` element.


lpil:
Thanks!

PR #543, opened 2015-03-13
Add SystemVerilog language snippets

BooneJS:
Hi,

I've created a snippets file containing useful snippets for SystemVerilog. I believe I've followed the proper convention where appropriate.

Regards,
boonejs


lpil:
Hi BooneJS. Thanks for these

It's preferred that snippets that don't use any special ultisnips features use the snipmate format, so everyone can use them. Could you convert them please?

Also, the commits don't seem to be under an email added to your github account. It doesn't really matter, but if you'd like them to be connected you should add that email :)

BooneJS:
Hi,

I've added/confirmed the email address used to make the commits.

I admit that UltiSnips is the first vim snippets utility I've used. I've looked at vim-snippets' snipmate directory, and it appears that the only difference is the use of bare $0 vs. ${0}. Can you confirm that's how I make them snipmate compatible?

Thanks,
BooneJS

lpil:
Snippets are indented by one tabstop, and there is no closing delimiter. It's like python ;)

BooneJS:
Hi! I've updated the snippets, please take another look.

lpil:
Thank you :)

PR #548, opened 2015-03-24
bug in class's ending scope

BooneJS:
Hi,

I discovered and fixed a bug in the 'cl' shortcut's endscope.

Thanks,
BooneJS


SirVer:
Thanks!

PR #550, opened 2015-03-24
cs and sh

switch87:
Small changes in some snippets for cs like used in visual studio.

added GPL2 license to sh snippets

### DIFF
--- a/snippets/cs.snippets
+++ b/snippets/cs.snippets
@@ -71,7 +71,7 @@ snippet simc
 	}
 # if condition
 snippet if
-	if (${1}) {
+	if (${1:True}) {
 		${0}
 	}
 snippet el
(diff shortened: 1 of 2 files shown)


snippets/cs.snippets
```diff
@@ -103,15 +103,15 @@ snippet for
 	}
 # foreach
 snippet fore
-	foreach (var ${1:entry} in ${2}) {
+	foreach (${1:var} ${2:entry} in ${3}) {
```
lpil: Is it that you want to delete the `var` word sometimes? If so, can you also put the space in the `{}`s please.

lpil:
The sh snippets have nothing to do with sh. Could you remove them please?

snippets/cs.snippets
switch87: no, it is becouse you can use var or specify the variable type, it does not work without it.
like I wrote it it works the same as it does in visual studio.

switch87:
ok, done that.

lpil:
Thanks.

PR #553, opened 2015-04-01
defs snippet name clash

Sathors:
In UltiSnips/python.snippets defs is used to define a staticmethod, in snippets/python.snippets defs is used to define a method.

defs to define a method is not a good name, defm would be better.


SirVer:
Make sense to me. Thanks.

PR #554, opened 2015-04-02
Use double quotes rather than single quotes for strings.

majjoha:
Ruby strings with single quotes do not allow string interpolation which can be slightly annoying when using these snippets, so I thought it would be worthwhile to change the single quotes to double quotes instead to fix this issue.

### DIFF
--- a/snippets/ruby.snippets
+++ b/snippets/ruby.snippets
@@ -31,9 +31,9 @@ snippet beg
 	end
 
 snippet req require
-	require '${1}'
+	require "${1}"
 snippet reqr
-	require_relative '${1}'
+	require_relative "${1}"
 snippet #
 	# =>
 snippet end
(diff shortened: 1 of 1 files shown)


lpil:
It's common in the ruby community to not use double quotes unless you specifically need the features offered by them. As a result, I think changing to double by default would be annoying for most people.

https://github.com/bbatsov/ruby-style-guide#strings

For changing delimiters, I recomment [tpope's vim-surround](https://github.com/tpope/vim-surround). Once installed, you can change from single to double quotes with `cs'"`.

majjoha:
I had an idea it would be difficult to change it after so many years, but I thought I'd try it, anyway.

SirVer:
This is not really about habit, it is mostly about security and performance. Only use templating strings if you really need them, not as a default.

lpil:
I never really thought about that side of it. Good point :)

taq:
There's another point about using always double quotes: the no need to think if you'll be using single or double. Single won't work with interpolation, double will work with all the cases. Anyway, the language default way (as tools like Rubocop can check) is using single, so I think we can keep this way.

lpil:
Looks like the performance side is nonsense.

https://viget.com/extend/just-use-double-quoted-ruby-strings

PR #560, opened 2015-04-13
python: docstring: do not add a newline after '"""'

blueyed:
Ref: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings


lpil:
Thanks :)

Issue #559, opened 2015-04-10
C snippets do not follow the same style.


chavezgu:
The snippets in the snippets/ folder have a slightly different style that the ones in UltiSnipts, regarding the curly brace. The former have the open curly brace on the same line as the "for" or "if" statement, while the later have it below the line.

lpil:
Thank you for this issue. This is clearly not ideal.

Which style is dominant in the C community? How do we decide which to adopt?

AssailantLF:
I feel like the K&R style (open brace on the same line) are used more often, just because of the popularity and history of the book, not to mention the linux kernel coding standard. However, it would fit better with the popular plugin vim-surround to use the Allman style, since that's how it surrounds lines of code (open brace on its own line).

It's really tough since the differences are almost entirely aesthetic and trivial. I personally vote for the Allman style. If only there were some way to have a compromise. :/

chavezgu:
I would personally vote for the K&R style, for the same reasons @AssailantLF said. However, I was
just pointing out that is not ideal to have two different styles for the same language. Right now engines using "UltiSnips" folder use the K&R style meanwhile the "snippets" folder use the Allman.

I also agree that the differences are trivial, but I think right now we have the worst of both worlds by having both styles.

SirVer:
+1 for K&R style.

jalcine:
Since #562 is closed, wouldn't that close this issue? Just curious.

SirVer:
yup, I agree.

PR #562, opened 2015-04-14
Change for, if, else, and struct to K&R style.

chavezgu:
As stated in issue https://github.com/honza/vim-snippets/issues/559, some snippets in the UltiSnips folder were not following the K&R style contrary to the snippets folder. One could actually achieve an inconsistent coding style while using snippets from both folders, e.g. if using the snippet "for" from UltiSnips and "forr" from snippets.

Hope this pull request helps to solve that.


lpil:
Thank you. :)

PR #557, opened 2015-04-18
Slightly smarter Ultisnips snippet file snippets

benhamill:
What would folks think about a PR that changed the Ultisnips snippet for Ultisnips snippet files that triggers on `snip` (what?) so that the description was optional and whitespace was trimmed from the end of the first line (in the case that you don't want any options)?

### DIFF
--- a/UltiSnips/snippets.snippets
+++ b/UltiSnips/snippets.snippets
@@ -2,8 +2,8 @@ priority -50
 
 # We use a little hack so that the snippet is expanded
 # and parsed correctly
-snippet snip "Snippet definition" b
-`!p snip.rv = "snippet"` ${1:Tab_trigger} "${2:Description}" ${3:b}
+snippet snip "Snippet definition" bs
+`!p snip.rv = "snippet"` ${1:Tab_trigger}${2: "${3:Description}" ${4:b}}
 $0
 `!p snip.rv = "endsnippet"`
 endsnippet


benhamill:
I've turned this into a Pull Request. Can you re-open for review and merge?

benhamill:
@SirVer Alert. Meant to mention you in the above comment. Sorry.

UltiSnips/snippets.snippets
```diff
@@ -2,8 +2,8 @@ priority -50
 
 # We use a little hack so that the snippet is expanded
 # and parsed correctly
-snippet snip "Snippet definition" b
-`!p snip.rv = "snippet"` ${1:Tab_trigger} "${2:Description}" ${3:b}
+snippet snip "Snippet definition" bs
```
SirVer: that is too generic. It strips whitespaces from all lines - but there might be significant whitespaces in the contents of the snippet at the end. You have to special case the first line.

benhamill: Ah. Good catch. I don't think this is actually required, since the second tabstop encompasses the space between 3 and 4. Should I just remove the `s`?

SirVer:
Reviewed files:
- UltiSnips/snippets.snippets @ r1

---

<sup>**[UltiSnips/snippets.snippets, line 5 \[r1\]](https://reviewable.io:443/reviews/honza/vim-snippets/557#-JnZqfId475CiOlPO18L-r1-5)** ([raw file](https://github.com/honza/vim-snippets/blob/a021bfa95ef79198f785645b8b79a8c3ebe39348/UltiSnips/snippets.snippets#L5)):</sup>
No, I think the line should look something like this:

~~~
`!p snip.rv = "snippet"` ${1:Tab_trigger}`snip.rv = " " it t[2] else ""`"${2:Description}"`snip.rv = " " it t[3] else ""`${3:b}
~~~

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/honza/vim-snippets/557)

SirVer:
@benhamill ping?

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/honza/vim-snippets/557)

benhamill:
Sorry. Life became a thing. I couldn't immediately parse your comment, so I need to look more at it when things settle down. It looks like you took out the tab stop I added? But maybe I'm just reading it wrong. I'm not super familiar with the advanced stuff.

SirVer:
I only removed the spanning snippet $2 - I deemed it not really that necessary. But I do not mind having it. 

What I added is conditional whitespaces for each trigger - so that there is no space after a trigger when the trigger has no length.

benhamill:
Oh, I see. Yeah. My main thing was just that, if I don't want to have a description or options, I want to be able to just hit delete and tab and move on, rather than have empty quotes in there. Maybe the snippet is meant to encourage descriptions on everything?

SirVer:
descriptions on snippets are definitively encouraged. But user ergonomics are more important. I do not feel strongly about the extra tabstop you had - feel free to add it back in it. But I still think the empty space can be conditional.

SirVer:
The 's' option must still be removed, it might murder some snippets otherwise.

SirVer:
Closing for inactivity.

PR #565, opened 2015-04-18
Enhanced Ultisnips Elixir snippets

benhamill:
Many of these collapse several snippets Elixir snippets into a single Ultisnip
Elixir snippet. Notably, snippets for `if` and `unless` don't have
else-including companions, but, rather, the included `end` is an optional final
section and there's an `else` snippet.

### DIFF
--- a/UltiSnips/elixir.snippets
+++ b/UltiSnips/elixir.snippets
@@ -1,0 +1,126 @@
+snippet if
+if ${1:condition} do
+	${2:expression}
+${3:end}
+endsnippet
+
+snippet unless
+unless ${1:condition} do
+	${2:expression}
+${3:end}
+endsnippet
+
+snippet else
+else
+	${1:expression}
+end
+endsnippet
+
+snippet iff
+if ${1:condition}, do: ${2:expression}${3:, else: ${4:expression}}
+endsnippet
+
+snippet unlesss
+unless ${1:condition}, do: ${2:expression}${3:, else: ${4:expression}}
+endsnippet
+
+snippet def
+def ${1:name}${2:(${3:parameters})}${4: when ${5:guard-condition}} do
+	${0:expression}
+end
+endsnippet
+
+snippet defp
+defp ${1:name}${2:(${3:parameters})}${4: when ${5:guard-condition}} do
+	${0:expression}
+end
+endsnippet
+
+snippet deff
+def ${1:name}${2:(${3:parameters})}, do: ${0:expression}
+endsnippet
+
+snippet defpp
+defp ${1:name}${2:(${3:parameters})}, do: ${0:expression}
+endsnippet
+
+snippet fn "inline function" s
+fn ${1:${2:args} }-> ${0:expression} end
+endsnippet
+
+snippet for
+for ${1:item} <- ${2:items}${3:, into: ${4:%{}}} do
+	${0:expression}
+end
+endsnippet
+
+snippet cond "cond do" s
+cond do
+	${1:pattern match} -> ${2:expression}${3:
+	->$4}
+end${0}
+endsnippet
+
+snippet rec "receive do" s
+receive do
+	${1:pattern match} -> ${2:expression}${3:
+	->$4}
+end${0}
+endsnippet
+
+snippet case "case do" s
+case ${1:expression} do
+	${2:pattern match} -> ${3:expression}${4:
+	->$5}
+end${0}
+endsnippet
+
+snippet try
+try do
+	${1:expression}
+rescue
+	${2:pattern match} -> ${3:expression}${4:
+	->$5}
+end${0}
+endsnippet
+
+snippet -> "case/cond pattern match" s
+${1:pattern match} -> ${2:expression}${3:
+->$4}
+endsnippet
+
+snippet @doc
+@doc """
+${1:A really good excuse for the code that follows.}
+"""
+def${0}
+endsnippet
+
+snippet @mod
+@moduledoc """
+${1:A really good excuse for this entire module.}
+"""
+endsnippet
+
+snippet req
+require ${1:Module}${2:, as: ${3:NewName}}
+endsnippet
+
+snippet ali
+alias ${1:Module}${2:, as: ${3:NewName}}
+endsnippet
+
+# To encourage the use or :only over :except.
+snippet imp
+import ${1:Module}${2:, only: [${3:function}: ${4:arity}]}
+endsnippet
+
+snippet test
+test "${1:behaves in some way}"${2:, %{${3:test_context}: ${4:dictionary}}} do
+	${0:test case}
+end
+endsnippet
+
+snippet pry
+require IEx; IEx.pry
+endsnippet
--- a/snippets/elixir.snippets
+++ b/snippets/elixir.snippets
@@ -1,81 +1,81 @@
 snippet do
 	do
-		${0}
+		${0:expression}
 	end
 snippet if if .. do .. end
-	if ${1} do
-		${0}
+	if ${1:condition} do
+		${0:expression}
 	end
 snippet if: if .. do: ..
 	if ${1:condition}, do: ${0}
 snippet ife if .. do .. else .. end
 	if ${1:condition} do
-		${2}
+		${2:expression}
 	else
-		${0}
+		${0:expression}
 	end
 snippet ife: if .. do: .. else:
-	if ${1:condition}, do: ${2}, else: ${0}
+	if ${1:condition}, do: ${2:expression}, else: ${0:expression}
 snippet unless unless .. do .. end
-	unless ${1} do
-		${0}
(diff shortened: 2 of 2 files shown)


UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
...
+snippet for
+for ${1:item} <- ${2:items}${3:, into: ${4:%{}}} do
+	${0:expression}
+end
+endsnippet
+
+snippet cond "cond do" s
+cond do
+	${1:pattern match} -> ${2:expression}${3:
+	->$4}
```
benhamill: The idea is that, here, if you wanted another pattern/expression set, you could just hit `<tab>` and trigger the `->` snippet, so it chains as long as you need more.

UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
+snippet if
+if ${1:condition} do
+	${2:expression}
+${3:end}
```
benhamill: If you wanted an if/else, you can just replace `end` with `else` and hit `<tab>` to get it.

benhamill:
I'd love some feedback on some of the conventions I used here. For instance, I used a full word (e.g. `if`) for the multi-line version of an expression and then doubled the last letter of the word (e.g. `iff`) for the single-line version of the expression.

UltiSnips/elixir.snippets
lpil: I don't like this because it deviates from the standard, and
```
if<tab>true<tab>foo<tab>else<tab>foo
```
doesn't seem any better than
```
if<tab>true<tab>foo<CR>else<CR>foo
```
And when you don't want an else it's actually worse.

UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
...
+endsnippet
+
+snippet else
+else
+	${1:expression}
+end
+endsnippet
+
+snippet iff
```
lpil: We're already using `if:` in the other format, why deviate here? :)

snippets/elixir.snippets
```diff
@@ -1,81 +1,81 @@
 snippet do
 	do
-		${0}
+		${0:expression}
```
lpil: As in the README, please avoid using/adding useless placeholder text.

lpil:
Hey! :)

You've introduced a lot of duplication here, we'd like to avoid this where possible.

It's also pretty hard to digest the changes in such a large pull request- there's a lot of things to talk about here.

UltiSnips/elixir.snippets
lpil: That's an interesting idea. I'll have to give it a try when I have time and see how it feels. :)

UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
...
+end${0}
+endsnippet
+
+snippet -> "case/cond pattern match" s
+${1:pattern match} -> ${2:expression}${3:
+->$4}
+endsnippet
+
+snippet @doc
+@doc """
```
lpil: This really doesn't need to be an ultisnips snippet

UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
...
+snippet ali
+alias ${1:Module}${2:, as: ${3:NewName}}
+endsnippet
+
+# To encourage the use or :only over :except.
+snippet imp
+import ${1:Module}${2:, only: [${3:function}: ${4:arity}]}
+endsnippet
+
+snippet test
```
lpil: I prefer not having so many optional groups to be removed from a common snippet like this as it means you have to stop and think about what you want the snippet to do. Instead I'd rather have multiple snippets, one for each common test declaration, as currently is with `test` and `testa`

`testa` could probably be improved though.

UltiSnips/elixir.snippets
```diff
@@ -0,0 +1,126 @@
...
+require ${1:Module}${2:, as: ${3:NewName}}
+endsnippet
+
+snippet ali
+alias ${1:Module}${2:, as: ${3:NewName}}
+endsnippet
+
+# To encourage the use or :only over :except.
+snippet imp
+import ${1:Module}${2:, only: [${3:function}: ${4:arity}]}
```
lpil: I'd rather merge 3 and 4. It saves one keypress at the moment, not worth the mental overhead.

UltiSnips/elixir.snippets
benhamill: movement distance. Inspired by `yy` to copy a line in vim and other commands like it: Your finger's already on the last letter.

UltiSnips/elixir.snippets
benhamill: I wasn't optimizing for keystrokes, here, but for a reduced number of things to remember. I think you have a point, though... If there was a potential `elsif` or similar, that be that would make more sense.

Is there a standard beyond "what's been done before in this repo"? Not that that's necessarily a poor standard, I'm just wondering if I'm missing a larger context or something.

UltiSnips/elixir.snippets
benhamill: You can start the snippet and basically just hit `<tab>` forever and it keeps going.

snippets/elixir.snippets
benhamill: A. I missed that. I'll undo _all_ my changes to elixir.snippets changes, then. I thought it was nicer this way, but project standards are a thing. Thanks.

UltiSnips/elixir.snippets
benhamill: Good call. I'll pull this out and make a new PR to change the one in elixir.snippets, then? I felt like this PR would be easier to discuss if it were all additive.

UltiSnips/elixir.snippets
benhamill: I was thinking it saved two (the `:` and the space), but I'm not super attached to it. What do you mean by "it doesn't work for any functions other than the first one you name"? I thought you could name any number of functions in the `only:` clause... Am I wrong about that or misunderstanding you?

UltiSnips/elixir.snippets
benhamill: Well, if you don't want the context dictionary, you delete it with a single keypress.

Test with context:
```
test<tab>solves all my problems<tab><tab>logged_in_user<tab>logged_in_user<tab>some test code
```

Test without context:
```
test<tab>solves all my problems<tab><delete><tab>some test code
```

Again, I was trying to optimize for fewer snippet triggers to remember and felt like wanting the context dictionary was probably the more common case. In this case, `testa` would be 1 keystroke shorter (less a `<tab>` and the `<delete>` and plus the `a`. To me, the mental burden of things-to-remember is more important than a few keystrokes.

If others disagree, that's cool. Whatever doesn't get merged can stay in my personal snippet file.

UltiSnips/elixir.snippets
lpil: I prefer not to bind things to hitting the same key twice in a row, for some reason I find this harder to type than two different keys.

I do think this is better though- `:` requires two keys to type.

Both the snippets format and ultisnips format snippets should have the same trigger though.

UltiSnips/elixir.snippets
lpil: I think remembering that `end` is a group is more to remember than that you can type `else` at any time. I could imagine tying a larger block of code inside the `if` and then stumbling over the fact the `end` has been selected when I press tab later.

Pretty much just what's already been made. I've been trying to clean up some of the languages, but honestly there's still rather a lot of mess.

UltiSnips/elixir.snippets
lpil: Sounds good. Could you either push another commit to this branch, or edit this commit to remove these ones that could be in the snippet format please?

UltiSnips/elixir.snippets
lpil: As in I type the first one with `foo<tab>3`, but if I need another I have to type `,<CR>bar/2`, rather than `,bar<tab>3`.

snippets/elixir.snippets
lpil: It's good in some places, but when everything in a language is an expression, saying that something's an expression doesn't mean much :)

UltiSnips/elixir.snippets
lpil: > To me, the mental burden of things-to-remember is more important than a few keystrokes.

I agree, but I feel the mental burden of complicated snippets is harder than that of just typing a second argument. The problem with snippets like these in my opinion is that while you're working with one you have to go through a mental loop of looking at it, registering where you are, deciding what action to take, and repeat.

Still, it's not all about me. Perhaps this is what we want. Does anyone else have an opinion? What do we do in other languages?

SirVer:
Guys, is this still progressing?

PR #568, opened 2015-04-23
fixed $this->table attribute access in ci_model_crudl and removed function doc comments

rtorralba:


SirVer:
This branch contains more than the advertised fix. Please correct that.

rtorralba:
I updated it, just fixed $this->table attribute access and removed doc comments because i think that make the code more unreadable

SirVer:
Thanks. Merged. I think removing the doc snippets here is a reasonable choice to make - they are rather heavyweight and opinionated.

PR #571, opened 2015-04-26
Made style for C and C++ snippets more consistent

cparadis6191:
The style is obviously highly personal but it makes more sense to pick one and stick with it.
All brackets are now on the same line.
Parentheses surrounding conditional statements and arguments (functions, if, while, switch)
aren't padded with a space anymore.


cparadis6191:
I guess this should be closed as it actually does follow the agreed upon K&R style (which I recently learned about).

lpil:
Thank you :)

SirVer:
sorry I merged something against the agreed upon style earlier this week - I thought we went for the google style in our discussion and remembered incorrectly.

PR #573, opened 2015-04-27
Elixir: Improve indentation of rec snippet

blacksails:
I think this is the proper way to indent receive blocks.


lpil:
The official vim plugin agrees with you. Thanks!

PR #574, opened 2015-04-28
Fixed ->table attribute access

rtorralba:


SirVer:
Gotta trust you on that - never heard of codeigniter. Thanks.

Issue #468, opened 2014-10-09
Allow more control over loading of Javascript snippets


lpil:
Hello all!

So lately I've been having a problem with the Javascript snippets. There's a great deal of library specific snippets (namely jQuery ones) that I'm uninterested in using, yet with the current directory structure there doesn't seem to be a way to prevent them from being loaded (as least as far as I can tell).

The issue is that because the directory is called Javascript all the contents will be loaded by vim-snipmate as they match `scope/*.snippets`, and if I change the scope from 'javascript' I lose the vanilla JS snippets too.

My first thought was to remove the `javascript/` directory and call them `javascript_jquery.snippets` etc in the parent dir, but this will have the same problem, as it also uses the search pattern `scope_*.snippets`.
If we used filenames that don't match either of these suddenly we have snippets that were loaded by default that no longer are, and I expect this would annoy users, so this probably isn't a good idea overall.

How do you deal with situations like these?

Cheers,
Louis

mmwtsn:
From SirVer's above answer it sounds like creating a `UltiSnips` directory in `~/.vim/` and adding a snippet files there will give you the ability to add, remove, or modify snippets without having to modify the plugin files directly and risk losing your changes whenever the plugin is updated.

Is this documented anywhere?

I would like to disable the JavaScript Jasmine snippet file as [a snippet within that file](https://github.com/honza/vim-snippets/blob/b11832aa686e7cc18385110398ca4e8b5688b5d9/UltiSnips/javascript_jasmine.snippets#L45-L46) is conflicting with [a Node.js snippet](https://github.com/honza/vim-snippets/blob/4496631cf19dde58f3fc95b573637464ac1b7698/snippets/javascript/javascript.node.snippets#L4-L5) I use regularly.

Being able to opt in or out of individual snippets (perhaps by turning them off one by one in your `.vimrc` or turning off an entire language then explicitly requiring only the snippets you're interested in) would be great.

lpil:
What an obnoxious set of snippets for a testing framework. One has a single letter trigger!

I'll open a PR to remove a bunch of these.

SirVer:
> Is this documented anywhere?

Sure, in UltiSnips docs. 

> Being able to opt in or out of individual snippets (perhaps by turning them off one by one in your .vimrc or turning off an entire language then explicitly requiring only the snippets you're interested in) would be great.

[...]

PR #580, opened 2015-05-17
Remove some JS jasmine snippets

lpil:
There's quite a lot of removal here. What do you think?

https://github.com/honza/vim-snippets/issues/468#issuecomment-102473202

### DIFF
--- a/UltiSnips/javascript_jasmine.snippets
+++ b/UltiSnips/javascript_jasmine.snippets
@@ -1,184 +1,93 @@
 priority -50
 
-# JavaScript versions -- from the TextMate bundle + some additions
-# for jasmine-jquery matchers
-#
-
-snippet des "Describe (js)" b
+snippet desc "Describe" b
 describe('${1:description}', function() {
 	$0
 });
 endsnippet
 
-snippet it "it (js)" b
+snippet it "it" b
 it('${1:description}', function() {
 	$0
 });
 endsnippet
 
-snippet bef "before each (js)" b
+snippet before "before each" b
 beforeEach(function() {
 	$0
 });
 endsnippet
 
-snippet aft "after each (js)" b
+snippet after "after each" b
 afterEach(function() {
 	$0
 });
 endsnippet
 
-snippet any "any (js)" b
-jasmine.any($1)
-endsnippet
-
-snippet ru "runs (js)" b
+snippet runs "runs" b
 runs(function() {
 	$0
 });
 endsnippet
 
-snippet wa "waits (js)" b
-waits($1);
+snippet waits "waits" b
+waits(function() {
+	$0
+});
 endsnippet
 
-snippet ex "expect (js)" b
+snippet exp "expect" b
 expect(${1:target})$0;
 endsnippet
 
-snippet ee "expect to equal (js)" b
+snippet expe "expect to equal" b
 expect(${1:target}).toEqual(${2:value});
 endsnippet
 
-snippet eb "expect to be (js)" b
+snippet expb "expect to be" b
 expect(${1:target}).toBe(${2:value});
 endsnippet
 
-snippet em "expect to match (js)" b
+snippet expm "expect to match" b
 expect(${1:target}).toMatch(${2:pattern});
 endsnippet
 
-snippet eha "expect to have attribute (js)" b
+snippet expha "expect to have attribute" b
 expect(${1:target}).toHaveAttr('${2:attr}'${3:, '${4:value}'});
 endsnippet
 
-snippet et "expect to be truthy (js)" b
+snippet expt "expect to be truthy" b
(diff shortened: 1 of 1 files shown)


UltiSnips/javascript_jasmine.snippets
```diff
@@ -1,10 +1,6 @@
 priority -50
 
-# JavaScript versions -- from the TextMate bundle + some additions
-# for jasmine-jquery matchers
-#
-
-snippet des "Describe (js)" b
+snippet desc "Describe (js)" b
```
SirVer: use `describe` as trigger

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
 });
 endsnippet
 
-snippet bef "before each (js)" b
```
SirVer: this one is very useful. `before` is an acceptable trigger though

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
...
-beforeEach(function() {
-	$0
-});
-endsnippet
-
-snippet aft "after each (js)" b
```
SirVer: same here

SirVer:
a lot of these snippets are very useful to me on a daily basis. How about we just pull them out of javascript into another filetype?

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
...
-afterEach(function() {
-	$0
-});
-endsnippet
-
-snippet any "any (js)" b
-jasmine.any($1)
-endsnippet
-
-snippet ru "runs (js)" b
```
SirVer: `runs` is a better trigger, but that is useful.

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
...
-runs(function() {
-	$0
-});
-endsnippet
-
-snippet wa "waits (js)" b
```
SirVer: same here, but 'waits'

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
...
-waits($1);
-endsnippet
-
-snippet ex "expect (js)" b
```
SirVer: they are used very often in jasmine tests - I approve that the triggers are short. `ee` makes more sense for expect to equal than expe, so I do not agree your suggested change is ergonomical.

UltiSnips/javascript_jasmine.snippets
```diff
@@ -16,169 +12,58 @@ it('${1:description}', function() {
...
-snippet en "expect to be null (js)" b
-expect(${1:target}).toBeNull();
-endsnippet
-
-snippet ec "expect to contain (js)" b
+snippet expc "expect to contain (js)" b
 expect(${1:target}).toContain(${2:value});
 endsnippet
 
-snippet ev "expect to be visible (js)" b
```
SirVer: I think it is fine to remove these - they are rather obscure.

SirVer:
mmh.... those snippets *are* super useful. I hate to see them go. Also, short triggers for common matchers seem fine to me. Why do you want to cut them down?

UltiSnips/javascript_jasmine.snippets
lpil: `desc` is the trigger used in other popular BDD frameworks, I think it's probably better to keep things consistent, as recommended in the readme.

UltiSnips/javascript_jasmine.snippets
lpil: I disgree `ee` is far too short a trigger for this.

Other BDD testing framework snippets I've looked at don't have many snippets like these, and the few that do have triggers like the one used here, or longer.

lpil:
Because the Javascript namespace is a mess of library specific snippets and these often clash.

Additionally they don't match many of the other BDD test framework triggers.

If we want to keep more of them I say we move them (and the other JS library snippets) out of the default JS namespace

SirVer:
> If we want to keep more of them I say we move them (and the other JS library snippets) out of the default JS namespace

I think that is a good approach to take. Let's teach people how to include the ones they want, i.e. using `:UltiSnipsAddFiletypes`.

mmwtsn:
+1 for moving these out of the JavaScript namespace and letting people add them in as needed.

If we clear up the JavaScript namespace we'll be able to support more testing libraries—Qunit, Mocha, Chai's various assertion styles, etc.—using sane triggers (e.g. `before`) without collision.

Issue #370, opened 2014-05-25
removing angularjs snippets.


RyanPineo:
When I first made the pull request to add a few angularjs snippets it was unclear if it would be a good fit to put them in this repo.  The reason being that there is no way to distinguish a file using angularjs from any other javascript.

There are quite a lot of snippets that could be included for use with angularjs and I think because of that and the above reason they should be moved to their own repo.

In fact there is already a good repo for such snippets at https://github.com/matthewsimo/angular-vim-snippets.

Just my 2 cents.

alexlafroscia:
Did a solution to the JS framework snippets ever get worked out?  The Ember snippets are pretty out-of-date, since the `App` global object really isn't a thing anymore.

lpil:
Sadly not. I think we're currently leaning towards removing them from the Javascript namespace.

If the Ember snippets are out of date we should update or remove them.

mmwtsn:
Perhaps we could open a few issues for updating the various JavaScript vendor snippets that are known to be outdated or lacking and assign a `needs-pr` label to them. I've found that's an easy way to flag low-hanging fruit for new contributors. Once the `javascript-vendor` directory is set up someone familiar with the framework or library should be able to jump in.

"Update the Ember.js snippets to version x.xx"
"Update the Angular.js snippets to version x.xx"

Et cetera.

alexlafroscia:
@lpil I'm working on a larger-scale plugin from Ember development that will include plugins that are more appropriate for the Ember CLI.  Maybe it would make more sense to just remove them from here and direct people to another source for more specific snippets?

lpil:
It would be nice to keep them in here as this repo is supposed to be a one-stop-shop for snippets. The problem is mainly with not polluting the top level namespaces with library snippets.

alexlafroscia:
Is there a way for turn off library snippets? For example, I don't ever want any of the Angular snippets in my Ember projects.  Is it just a naming convention that ensures they aren't visible by accident?

lpil:
It's purely the naming convention.

We should come up with a new one for snippets that belong to a format but are not to be loaded by default.

mmwtsn:
How about this?

- Base language snippets are available by default (e.g. `javascript.snippets`)
- Additional library, framework, or legacy version snippets are available in a `<language>-vendor` directory (e.g. `javascript-vendor/angular.snippets`)
- Vendor snippets are not available by default but can be included by the user on a snippet-by-snippet basis as needed

[...]

PR #581, opened 2015-05-21
Move UltiSnips vendor snippets out of default namespace

mmwtsn:
This is a quick but breaking change to move framework, library, and other external snippets into vendor directories as discussed in #370.

According to the UltiSnips documentation at `UltiSnips-adding-snippets` these directories will not be sourced automatically as only directories with the exact filetype will match (e.g. `javascript`).

Testing this locally that seems to be the case.

Snippets can be added back in using the `UltiSnipsAddFiletypes` command. For example, to restore the Ember.js snippets for all JavaScript files:

```vim
:UltiSnipsAddFiletypes javascript-vendor/ember.js
```

This should help to clear up the default namespace while giving users more control over what vendor snippets get included in their local environment. We can also support a greater variety of libraries here with potentially conflicting triggers (e.g different testing frameworks for a single language) which should help with the "one stop shop" idea behind this repository.


SirVer:
> `:UltiSnipsAddFiletypes javascript-vendor/ember.js`

This will not work. You can add additional filetypes to search, but they will still be searched the way `ftplugin` searches for files in `ftplugin`. You need to do this instead:

`:UltiSnipsAddFiletypes vendor-javascript-ember`

This will include `vendor-javascript-ember.snippets`,  `vendor-javascript-ember_*.snippets` and  `vendor-javascript-ember/*.snippets`. So you have these three choices for naming your files - the subdirectory approach you currently choose will not work to separate the different vendors properly.

lpil:
What's the situation here @mmwtsn? :)

mmwtsn:
Got caught up moving apartments :)

I revised the commit yesterday with namespaced file names rather than the directory approach as @SirVer recommended above. To keep snippets grouped alphabetically, I went with `*-vendor-*.snippets` which leaves room for additional language snippets that we mightn't want to include in the base file such as upcoming releases (e.g. ES6).

Using the branch locally everything seemed to be working as expected.

SirVer:
I think `-vendor-` can be dropped. `javascript-jasmine` is just as good IMHO. otherwise lgtm.

mmwtsn:
Thanks @SirVer. Updated with your suggestion. This should be good to go.

Thoughts on adding a note to the README if this is merged in? Current users might be wondering what happened to some of their snippets after they update. New users would benefit from knowing that there are extra snippets available if they're using UltiSnips.

SirVer:
Sure, documentation would be helpful :)

mmwtsn:
All set!

SirVer:
Sorry, this slipped through the cracks again :/. Could you fix merge conflicts?

mmwtsn:
Fixed.

SirVer:
Thanks!

lpil:
:dancer: Hooray

mmwtsn:
:tada: It's in!

PR #586, opened 2015-05-30
Add test for incorrect indentation, add Travis CI

lpil:
After finding yet more broken snippets today I've added a test for snipmate format snippets that are indented with spaces. It's maybe not the best way to do it (I don't speak shell), but it seems to work.

Coupled with Travis CI we will easily be able to tell if pull requests can this particular problem or not. @honza can you turn on travis CI for this repo? I presume you have to be the one that does it.

What do we think? A good addition to the repo?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: c
+sudo: false
+
+notifications:
+  email: false
+
+script:
+  - ./tests.sh
--- a/tests.sh
+++ b/tests.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+SPACED=$(grep -REn '^ .+' --include '*.snippets' snippets);
+
+if [[ -n SPACED ]]; then
+  echo These snippet lines are indented with spaces:;
+  echo;
+  echo "$SPACED";
+  echo;
+  echo Tests failed!
+  exit 1;
+fi
+
+echo Tests passed!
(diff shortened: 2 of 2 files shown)


lpil:
Also, should we do this for the utlisnips snippets too?

SirVer:
Great initiative! I was thinking about adding something along these lines too - but I did not get around to it. Thanks for doing this! I leave this for you to merge and activate on travis.

> Also, should we do this for the utlisnips snippets too?

No need I think. We never had invalid syntax snippets for UltiSnips yet, because UltiSnips is much more aggressiv about validating the syntax. And checking that that the snippet does what it should is also much harder with UltiSnips snippets as they can contain arbitrary logic. While nice I think that would be a big time investment that might not really pay off.

SirVer:
@honza ping, turns out you must flip the switch on travis ci to get this rolling.

lpil:
#591 wouldn't have been needed with this.

@honza Any thoughts?

honza:
@SirVer @lpil 

Sorry for the delay. I have enabled TravicCI for this repo. I didn't fill out any of the fields. Is that necessary?

honza:
TravisCI is now building on every push. I added the badge to the README. You can see the build outputs by clicking on it.

lpil:
Excellent, thank you!

PR #590, opened 2015-06-06
cpp extends c

daeyun:
Only tested using ultisnips. Depending on how other plugins parse snippet files, it might cause errors.


SirVer:
No, that is officially supported by vim-snippets these days. Should be fine.

PR #594, opened 2015-06-29
Some variables need default value

laxtiz:
so many snippets use `g:snips_email` and `g:snips_github` .
but, them not define.

who are newbie to this snippets package will get error when them expand some snippet use the variables above.

[garbas/vim-snipmate](https://github.com/garbas/vim-snipmate) has give `g:snips_author` a default value.
code is here: https://github.com/garbas/vim-snipmate/blob/master/plugin/snipMate.vim#L47-L49

[SirVer/ultisnips](https://github.com/SirVer/ultisnips) do it yet.
code is here: https://github.com/SirVer/ultisnips/blob/master/plugin/snipMate_compatibility.vim#L6-L8

so i add these code for `g:snips_*`

### DIFF
--- a/plugin/vimsnippets.vim
+++ b/plugin/vimsnippets.vim
@@ -3,6 +3,19 @@ if exists("b:done_vimsnippets")
 endif
 let b:done_vimsnippets = 1
 
+" Some variables need default value
+if !exists("g:snips_author")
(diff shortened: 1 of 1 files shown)


SirVer:
Wow, I thought we had them for a long time. Thanks!

PR #596, opened 2015-07-09
recommended python encoding declaration

tevino:
I guess this is the recommended form in Python 2 and Python 3.

https://docs.python.org/2/reference/lexical_analysis.html#encoding-declarations
https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations


SirVer:
Makes sense. Thanks.

PR #600, opened 2015-07-17
Javascript Postincrementers

nguyen-le:
Most of the time, I've seen the post-increment with no whitespace between:  ```i++``` instead of ```i ++```.


lpil:
That is weird- this is certainly an improvement. Thanks!

PR #601, opened 2015-07-23
UltiSnips/python: Honour quoting config for ifmain snippet

rsommer:
quoting style is already configurable, so it should be used where appropriate.


SirVer:
thanks.

PR #602, opened 2015-07-25
python: Allow to set triple quote style separately

audriusk:
I prefer using single quotes for all string literals except multiline strings, where I use double quotes. This change allows such style of quoting and doesn't change the behavior for those who use one style for all cases.


SirVer:
:+1:

PR #603, opened 2015-07-29
Add CommonJS snippet

rpoddighe:
Snippet for including CommonJS/Node.js-styled modules. Works for core and file modules! Generates idiomatic variable names by default when using hyphenated file names. Ex:

```
vreq<tab>./../models/some-model.js
```

expands to

```js
var SomeModel = require('./../models/some-model.js');
```

### DIFF
--- a/UltiSnips/javascript.snippets
+++ b/UltiSnips/javascript.snippets
@@ -162,4 +162,10 @@ snippet req "require an AMD module"
 require([${1:'dependencies'}], ${2:callback});
 endsnippet
 
+# CommonJS snippets
+
+snippet vreq "assign a CommonJS-style module to a var"
+var ${0:${1/(.+\/)*(\w+)(-|\b|$)(\..+$)?/\u$2/g}} = require('${1}');
+endsnippet
+
 # vim:ft=snippets:


lpil:
Hooray for JS modules!

Once the official JS module syntax becomes more commonplace we may want to move this out of the global namespace, but for now this seems OK.

rpoddighe:
There's still a lot to gain as to including Node snippets! For starters, we can move this to a separate file and add the Node snippets from the snipMate folder, but since UltiSnips can also parse snipMate snippets I'm not sure if it's desired to have duplicates.

lpil:
I'm not sure I entirely understand the loading system, so someone else might be better to talk about that.

rpoddighe:
It says [here](https://github.com/SirVer/ultisnips/blob/67fbdb2ad8c965378bdd60c892d3a43049caf3f1/doc/UltiSnips.txt#L366-L367) that the snipMate snippets folder is reserved and not loaded, so I guess that it shouldn't be a problem.

SirVer:
No, it says that the folder is reserved for snipmate snippets. But Ultisnips does load them - they are used in parallel with Ultisnips' own.

rpoddighe:
Oh, my mistake, I misinterpreted that. It seems though that in the case of duplicates, UltiSnips snippets take precedence over snipMate snippets, so #604 does not raise conflicts, which was my main concern.

PR #604, opened 2015-07-29
Move Node snippets to separate file

rpoddighe:
I moved the Node module snippet to a separate file and added some sensible snippets from the snipMate folder. I omitted the EventEmitter snippets because they conform to regular JS event snippets and are not Node-specific.


lpil:
This is great, thank you very much. :)

PR #610, opened 2015-08-14
Adding reactJS for coffeescript support

kursion:


lpil:
We probably don't want these in the default namespace, so the filename should be this

```
snippets/coffee-react.snippets
```

Is that correct, @SirVer ?

SirVer:
Yup, correct. And https://github.com/honza/vim-snippets#vendor-snippets applies.

kursion:
I've added this to `UltiSnips/coffee-react.coffee` and changed the structure a bit. I'm trying to activate it for me but it doesn't work.

I'm doing the following: `:UltiSnipsAddFiletypes coffee-react` but my snippets aren't shown :(

SirVer:
kursion, I cannot repro this. It works for me (using jasmine as example):

```
vim /tmp/blub.coffee
:set filetype=coffee
fun<expand> -> works, from coffee.snippets
des<expand> -> does not work, from coffee-jasmine.snippets
:UltiSnipsAddFiletypes coffee-jasmine<cr>
des<expand> -> now works.
~~~~

Where should we start debugging this?

kursion:
My bad, it's working. ;)

SirVer:
Is this then ready for merge? @lpil are you happy with the changes?

lpil:
I can't say I'm an expert, but it looks OK to me. Are you happy, @kursion ?

kursion:
@lpil Yes, I'm using it everyday and it's pretty cool :) Thanks for the merge.

lpil:
Hooray!

PR #611, opened 2015-08-17
fix case of 'Desc' attribute in puppet/yumrepo snippet

benningm:
Attribute 'Desc' must be changed to lowercase.


SirVer:
thanks.

Issue #613, opened 2015-08-23
Should stylus snippets auto-indent?


SevereOverfl0w:
All the stylus snippets seem to auto-indent on insertion.

I have auto-indent on, so I don't personally need this behaviour. I imagine most people using a snippet are already at the indentation level required for their snippet to be inserted at.

Would a PR be welcomed changing this?

lpil:
Sorry, I'm not sure what you mean. Could you give an example please?

SevereOverfl0w:
https://github.com/honza/vim-snippets/blob/master/snippets/stylus.snippets

Every line in this file is double indented (2 tabs).

When I expand a snippet with UltiSnips it inserts an indent for me Which means that this:
```
.post
  .something
    color red
    bgc<tab>
```
becomes:
```
.post
  .something
    color red
      background-color 
```

Notice that the line below `.something` indented further.

SevereOverfl0w:
This may have been accidental:
https://github.com/honza/vim-snippets/commit/e0fff1575abb816606c729751c8507c74e17c3cd
replaced 4 spaces with 2 tabs. Could have been intended to become 1.

lpil:
Looks like a mistake to me. Do you want to fix this, or shall I?

SevereOverfl0w:
I'll give it a go, why not!

PR #614, opened 2015-08-23
fix stylus indentation.

SevereOverfl0w:
Quickly ran:
`sed -i 's/^\t\t/\t/g'` against stylus.snippets, should fix issue https://github.com/honza/vim-snippets/issues/613


lpil:
Nice. Thanks :)

PR #615, opened 2015-08-25
javascript.snippets

dengyaolong:
add 
1. ce means console.err()
2. tryf means try/catch/finally
3. te means throw new Error('error message')

### DIFF
--- a/snippets/javascript/javascript.snippets
+++ b/snippets/javascript/javascript.snippets
@@ -242,6 +242,9 @@ snippet cl
 # console.debug
 snippet cd
 	console.debug(${0});
+# console.err
+snippet ce
+	console.err(${0});
 # console.trace
 snippet ct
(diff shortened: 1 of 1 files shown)


snippets/javascript/javascript.snippets
```diff
@@ -80,6 +80,18 @@ snippet try
...
+snippet tryf
+	try {
+		${1}
+	} catch (${2:e}) {
+		${0:/* handle error */}
+	} finally {
+		${3:/* be executed regardless of the try / catch result*/}
+	}
+# throw Error
+snippet te
```
lpil: I think te is a bit short.

dengyaolong: how about terr

lpil:
Looks good! :)

PR #616, opened 2015-08-27
Remove offensive haskell snippets

onemanstartup:
Haskell snippets won't show up in autocomplete because of this. 
https://github.com/Valloric/YouCompleteMe/issues/1214
It doesn't matter anyway. those snippets impossible to type.


lpil:
Seems reasonable. Thanks!

PR #619, opened 2015-08-31
Renaming UltiSnip files with hyphens to underscores.

jsdevel:


lpil:
This pulls all these snippets into the default namespace for those languages, we do not want to do this for framework or library specific snippets.

jsdevel:
Out of the box they're not picked up.  Is there documentation on activating them?

lpil:
That's the idea. :)

There is documentation in the README of this project, as well as the README of Ultisnips I believe.

https://github.com/honza/vim-snippets#vendor-snippets

jsdevel:
I see.  thanks for pointing it out in the docs!

PR #620, opened 2015-09-07
pep-3110

rurkowce:
Basically, the exception syntax has slightly changed in python3, and since according to [this](https://www.python.org/dev/peps/pep-3110/#compatibility) it has been backported, I see no harm in updating it


SirVer:
Thanks!

PR #624, opened 2015-09-23
Enhance python `with` snippet for UltiSnips

sophacles:
* the `with` snippet in UltiSnips now contains a VISUAL element
* The `as` keyword portion of a with statement now uses python
  interpolation to determine presence


SirVer:
really good changes. Thank you very much.

PR #626, opened 2015-09-30
add python style jedi

Ralph-Wang:
Args style like the picture. (learned from [https://leetcode.com](https://leetcode.com))

This style will make `jedi` knows the type of the arguments. 
otherwise, other style do not works.

Add it to python.snippets as a Choice.


SirVer:
beautiful! thanks.

PR #633, opened 2015-10-15
Replace - with _ to auto load UltiSnip snippets

ktilcu:
According to the UltiSnips Docs: any files that are to be auto-loaded
based on filetype need to follow a specific naming convention. Since
these snippets are in the UltiSnips folder it makes sense to follow that
convention.

> Using a strategy similar to how Vim detects |ftplugins|, UltiSnips
> iterates over the snippet definition directories looking for files
> with names of the following patterns: ft.snippets, ft_*.snippets, or
> ft/*, where "ft" is the 'filetype' of the current document and "*" is
> a shell-like wildcard matching any string including the empty string.


lpil:
This is deliberate. These library specific snippets should not pollute the global namespace.

PR #634, opened 2015-10-19
proposes to return to the convention ruby

qipoy:
in ruby 1.9+ ```=>``` or rocket hash, have been replacing with ```foo:```

```:foo => 'demo/index' ``` its same with ```foo: 'demo/index'```


lpil:
Great change, hooray! :)

PR #636, opened 2015-10-25
Improve Elixir defmo snippet with name based on filename

iurifq:
[asciicast](https://asciinema.org/a/2doetoamkact3qy4fb6tc09yw)

### DIFF
--- a/snippets/elixir.snippets
+++ b/snippets/elixir.snippets
@@ -70,7 +70,7 @@ snippet defma
 		${0}
 	end
 snippet defmo
-	defmodule ${1:module_name} do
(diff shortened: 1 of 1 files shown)


honza:
:+1:

lpil:
Sweet! Thanks :)

lpil:
A further improvement would be to take the path into account so `lib/foo/bar.ex` expands to `Foo.Bar` and `test/foo/bar_test.exs` expands to `Foo.BarTest` :)

iurifq:
Agreed :+1:

PR #637, opened 2015-10-29
add i and w option to eruby

mangege:
```
<b>=<b> # = not work
<b class="=">abc<b> # = not work
<b>lt<b> # lt not work
```


SirVer:
Great changes, thanks.

PR #638, opened 2015-11-04
Add comment ending ifndef block to ifndef c snippet

natemara:
It is generally considered good practice when defining C preprocessor if-statement blocks to include a closing comment so that the block will have the form:

    #ifndef _FOO_H_
    #define _FOO_H_
    ...
    #endif /* ifndef _FOO_H_ */

This is because preprocessor if-statements like this cannot be indented, so nesting can become messy. This should add those comments to the closing line of the `ifndef` snippet for the c language in both the Ultisnips and SnipMate formats.


lpil:
Great, thanks!

PR #639, opened 2015-11-06
Included Option pragma

epsilonhalbe:
- included OPTIONS_GHC pragma
- and removed unnecessary tab in module where clause


lpil:
Thanks!

PR #641, opened 2015-11-09
Cpp include snippets

dagon666:
Hi,

Since constantly removing the ".h" extension when working with C++ STL include files is pretty annoying I've decided to add a C++ specific version of "inc" snippet.

Additionally I've added a simple "inc" snippet for boost headers.


lpil:
Thanks :)

PR #642, opened 2015-11-10
tex snippets

kevinywlui:
Added two enum environment options.

Changed default text in lim to something that is likely more often used.

Added a snippet for set bracket.


honza:
Thanks! :sparkles:

PR #646, opened 2015-11-20
Fix overall svg canvas height/width attribute

humean:
Typically the appended svg on line 18-20 needs to have a width/height that includes the margins. Details about the convention can be found here: http://bl.ocks.org/mbostock/3019563


honza:
Thanks :sparkles:

PR #647, opened 2015-11-21
Add "ipy" snippet to embed IPython shell into the code

yevhen-m:
Hey, just added a new snippet "ipy" to be able to embed IPython shell into the code. What do you think?


lpil:
Thanks!

PR #648, opened 2015-11-26
Additions to the Lua snippet

ChickenNuggers:
I found the Lua snippet rather... lacking, and while I'm not too much an expert on writing things as large as the Python `class` or `def` expansion, this should at least be a bit helpful for starting off.


lpil:
>  I'm not too much an expert on writing things as large as the Python class or def expansion

Those snippets are voodoo ;)

Thanks!

PR #652, opened 2015-12-06
Get rid of closing slashes in every void element

mrzool:
Because is somebody really still writing HTML using XML syntax?

See [Sending XHTML as text/html Considered Harmful](http://hixie.ch/advocacy/xhtml) and [this](http://stackoverflow.com/questions/3558119/are-self-closing-tags-valid-in-html5) and [this](http://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br).

### DIFF
--- a/snippets/html.snippets
+++ b/snippets/html.snippets
@@ -159,18 +159,18 @@ snippet address
 		${0}
 	</address>
 snippet area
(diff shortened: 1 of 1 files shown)


lpil:
Thank you.

PR #653, opened 2015-12-07
ECMAScript 6 snippets.

brianvanburken:


lpil:
Could you remove the blank lines between snippets please?

Is cls a normal trigger for a class definition? I think cl or cla is more common.

Is interface an es6 feature?

brianvanburken:
I've removed the lines and, looking at other triggers, changed the class trigger to cla.
Also `interface` is not an ES6 feature. Don't know why I made that one 😅.

lpil:
Thanks! :)

PR #655, opened 2015-12-12
Add simpler arrow function snippet

lfilho:
Not only `af` is 1 less keystroke, and the keys are very close (home row of the keyboard). `=` + `shift` + `.` (`=>`) are too far away from each other and takes too long to type :)


lpil:
Nice. :)

lfilho:
That was fast! Thanks :)

PR #658, opened 2015-12-16
Fixed spacing issue with box and bbox.

yannack:
When commentstring had a trailing space, such as '# %s', the rendered
box was incorrect.


honza:
Thanks! :sparkles:

PR #662, opened 2015-12-30
Added Jinja2 snippets (copy of htmldjango.snippets)

jjpikoov:
Should be helpful for those who use for example this plugin:
    https://github.com/Glench/Vim-Jinja2-Syntax


lpil:
Is there a reason to keep the htmldjango snippets?

jjpikoov:
I don't think so.

lpil:
Would you like to remove them?

jjpikoov:
Yes, why not. Should I make new pull request?

lpil:
You can push another commit to this one :)

jjpikoov:
Done.

lpil:
Thanks!

somini:
I think this removal of htmldjango file is wrong. Jinja and the Django template language are similar, but different. Plus, the htmldjango filetype is detected automatically, suddenly my django snippets have stopped working.
https://docs.djangoproject.com/en/dev/topics/templates/
I think the best course of action is to keep both files with the same content, and diverge the jinja file as needed.

lpil:
If they are not the same language then this was a mistake. Thank you for letting us know.

somini:
Thanks for the quick revert.

lpil:
:)

PR #663, opened 2016-01-04
Convert cmake.snippets to lower case.

phcerdan:
As recommended in cmake.

Add snippets: ExternalProject (url and git), project, min, test
Rename snippet "cmake" to "init"


honza:
Thanks!

PR #664, opened 2016-01-06
Remove duplicate ruby block snippets

fuadsaud:
Defined here:

https://github.com/honza/vim-snippets/blob/master/snippets/ruby.snippets#L425


lpil:
Thank you!

PR #666, opened 2016-01-08
Args start at ${1}

crzidea:
${0} should be used as the end of snippet according to other snippets.
Besides, added new snippet `${` for argument of template strings. e.g.:
```
`${<tab>` // => `${cursor}`
```


lpil:
Great, thank you!

PR #667, opened 2016-01-10
sugest for simple modify snippet in node js

Pompeu:
node require(re) snippet now can use new types of variables not only var by defult, can change in $1


lpil:
Thank you

Pompeu:
thx u man

PR #668, opened 2016-01-11
PHP: trigger php 'array' snippet only for beginning of line.

phux:
currently the 'array' snippet is also triggered when you are in phpDoc blocks like

```
/**
 * @param array $foo
```


lpil:
Seems reasonable. I don't really understand this snippet though- why not just use `[]` ?

PR #672, opened 2016-01-16
Remove extra whitespace

kevinkoltz:
Prevents adding empty lines while using snipMate.vim


lpil:
Thanks for this, those spaces must have been annoying!

PR #674, opened 2016-01-29
python: add "bpdb" snippets for bpython

blueyed:


UltiSnips/python.snippets
```diff
@@ -634,6 +634,11 @@ snippet ipy "Embed IPython shell" b
 import IPython; IPython.embed()
 endsnippet
 
+snippet bpdb "Set bpython/bpdb breakpoint" b
+import bpdb
+bpdb.set_trace()
+endsnippet
```
lpil: No need to add it here if it is in the snipmate format already :)

blueyed: That applies to the others then also, right?
Is `snippets` / snipmate the preferred one?  I would have assumed that `UltiSnips` would be preferred?!

lpil: Snippets is the preferred as Ultisnips can read both formats :)

blueyed: But then snippets appears to have less features..?!

lpil: You're not using any of those features.

blueyed:
Closing in favour of #676.

PR #675, opened 2016-01-29
python: do not use multiple commands on a single line

blueyed:
Otherwise style checkers will usually complain about it, and e.g.
'flake8' cannot be made to ignore it with '# noqa' at the end.

### DIFF
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -627,19 +627,23 @@ endsnippet
 #####################
 
 snippet pdb "Set PDB breakpoint" b
-import pdb; pdb.set_trace()
+import pdb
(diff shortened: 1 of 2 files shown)


lpil:
Thank you.

Would you also like to remove the duplicated snippets from the ultisnips format?

blueyed:
> Would you also like to remove the duplicated snippets from the ultisnips format?

Ok.
But it's not clear to me which ones should be removed (see https://github.com/honza/vim-snippets/pull/674#discussion_r51286283).

lpil:
Commented in the other thread, thanks!

blueyed:
Closing in favour of #676.

PR #676, opened 2016-01-30
python: do not use multiple commands on a single line; add bpdb

blueyed:
Otherwise style checkers will usually complain about it, and e.g.
'flake8' cannot be made to ignore it with '# noqa' at the end.

I also adds the "bpdb" snippet, and only defines them as Snippets, not
UltiSnips.  Therefore "ipy" is only available as "iem" now.
Ref: https://github.com/honza/vim-snippets/pull/675#issuecomment-176840647

This supersedes #674 and #675.

### DIFF
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -626,22 +626,6 @@ endsnippet
 # Assertions & Tests #
 #####################
 
-snippet pdb "Set PDB breakpoint" b
-import pdb; pdb.set_trace()
-endsnippet
-
-snippet ipy "Embed IPython shell" b
-import IPython; IPython.embed()
-endsnippet
-
-snippet ipdb "Set IPDB breakpoint" b
-import ipdb; ipdb.set_trace()
-endsnippet
-
-snippet pudb "Set PUDB breakpoint" b
-import pudb; pudb.set_trace()
-endsnippet
-
 snippet ae "Assert equal" b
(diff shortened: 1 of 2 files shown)


lpil:
Thank you! :)

sloria:
-1 to this change. `pdb` / `ipdb` breakpoints should not be committed, so it doesn't matter that flake8 marks them as an error. In fact, I find it helpful for flake8 to mark the error.

Having breakpoints on a single line makes them more easily visible (due to the flake8 mark and the semicolon) and makes them easily delete-able.

Thoughts? Would the maintainers be open to reverting this?

lpil:
> In fact, I find it helpful for flake8 to mark the error.

To me this sounds like your linter should be reporting those statements as violations.

sloria:
@lpil That's just my point--my linter reported the statements as violations *before* this PR.

lpil:
And it should do afterwards. Open it as an issue in your linter.

blueyed:
> And it should do afterwards. Open it as an issue in your linter.

It is not a violation after this change anymore.

But I found it distracting to get reports on saving the file about this, which might even somehow mask real errors (if you're expecting to get an error/warning always).

sloria:
OK, I understand this is largely personal preference, and I don't want to push the issue. I'll just override these in my local snippets file.

PR #679, opened 2016-02-07
Change Rails snippet triggers

andyw8:
At present, `cla` maps to both Create Controller Class and Create
Functional Test class.

Using `cla` for creating a controller seems overly-general, since we may
want triggers for other types of classes, e.g. models. So I propose
changing this to `clac`.

And to remove ambiguity, I propose using `clact` for the controller test.


lpil:
Good call. :)

PR #683, opened 2016-02-12
Fixing nested snippet for Rust `let`/`letm`

kernelrain:
When expanding the `let` pattern, one would end up with something like 
`let I: I${4:type} = I;` 
(I representing the cursor jump marks), so the nested jump marks are inserted literally.

I fixed that unwanted behaviour by replacing the `let` and `letm` patterns by `let` (for immutable variable bindings with type inference), `lett` (for immutable variable bindings with explicit type annotation), `letm` (for mutable variable bindings with type inference) and `lettm` (for mutable variable bindings with explicit type annotation).

### DIFF
--- a/snippets/rust.snippets
+++ b/snippets/rust.snippets
@@ -34,11 +34,11 @@ snippet main "Main function"
 snippet let "let variable declaration with type inference"
 	let ${1}  = ${2};
 snippet lett "let variable declaration with explicit type annotation"
-	let ${1}: ${2}  = ${2};
+	let ${1}: ${2}  = ${3};
 snippet letm "let mut variable declaration with type inference"
 	let ${1}  = ${2};
 snippet lettm "let mut variable declaration with explicit type annotation"
-	let ${1}: ${2}  = ${2};
+	let ${1}: ${2}  = ${3};
 snippet pln "println!"
 	println!("${1}");
 snippet pln, "println! with format param"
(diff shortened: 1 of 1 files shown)


lpil:
I'm not sure I understand the problem, it seems to be working for me with Ultisnips. What plugin are you using?

kernelrain:
I'm using snipmate. I was able to reproduce the issue on two independent machines, that's why I assumed it was a bug.

lpil:
Ah yes, then it is. Thank you. Could you move the previous/current versions to the ultisnips format as they work with that plugin?

kernelrain:
I'm sorry but I'm not entirely sure what you mean. You said, the UltiSnips snippets appear to be working, that's why I only changed the snipmate snippets.

lpil:
You've removed the old ones in this commit, replacing them with new ones. The old ones work OK for ultisnips, so instead of replacing them entirely move them to the ultisnips directory. :)

kernelrain:
Maybe I'm wrong, but as far as I can tell, there is a dedicated `rust.snippets` file in the UltiSnips directory containing the previous definitions of `let`/`letm`

lpil:
Ah, in which case there's nothing to do. Thanks!

kernelrain:
Alright, thanks for merging :+1:

PR #684, opened 2016-02-13
the addition is a simple anonymous function snippet with a visual

bounceme:
tabstop and detection when following a non-word character


lpil:
Thanks, but this snippet already exists in the snipmate format.

PR #696, opened 2016-03-09
Rust: remove semicolons from constructor returns

Vzaa:
I'm sort of new to Rust and this is quite a minor change, so excuse me if I'm wrong here but I think the current constructor snippets are incorrect. With the semicolons the return type of the constructors are () not the struct type.


lpil:
Thank you :)

PR #697, opened 2016-03-10
Added Anonymous Function (javascript) for UltiSnips. Supports $VISUAL.

shepherdsam:
Also the `i` option is important in my use case. Which is writing callbacks...
```javascript
functionCall(af<tab>)
```
without it
```javascript
functionCall(<space>af<tab>)
```


lpil:
Thanks :)

PR #698, opened 2016-03-11
UltiSnips/java.snippets: Adds numbers to word in getArgs, Removes whitespace after braces

oddlydrawn:
For UltiSnips/java.snippets: currently variables added into clc parameters are ignored when they contain numbers, this PR changes it so numbers are allowed. Use case: [some classes have numbers in them](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/math/Vector2.html)

Braces also had a few characters of white space, which I've removed. It's easier to see it in a [screenshot with highlighted white space](https://i.sli.mg/VEAS8D.png) (left is current master, right is this PR's changes)

I don't think this breaks anything but haven't tested it extensively. I did try all 3 clc snippets and the md snippet and they look okay to me.

### DIFF
--- a/UltiSnips/java.snippets
+++ b/UltiSnips/java.snippets
@@ -124,7 +124,7 @@ for i in args:
 	\tthis." + i[1] + " = " + i[1] + ";\n\t}\n"
 
 	snip.rv += "\n\tpublic " + i[0] + " get" + camel(i[1]) + "() {\n\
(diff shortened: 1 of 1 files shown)


oddlydrawn:
Getters have an extra indent, fixing

lpil:
Thank you! Sorry this slipped through the net.

PR #700, opened 2016-03-15
Changed ifndef snippet to #ifndef

vulder:
This makes it consistent with the other C macro snippets, otherwise it was not possible to complete #ifndef.


lpil:
Thanks

PR #702, opened 2016-03-20
add snippets for three.js

yqrashawn:
I recently add some three.js snippets to folder ultisnips, I'm just getting in touch with three.js. So the snippets is not that good.


lpil:
Should these not be all one file?

yqrashawn:
Of course it should. I just conver it from other file and modified a little bit. 
My own version is already signal file and with a lot more snippets cause vim autocomplete plugin doesn't perform well with three.js. But it's with a lot Chinese comment. 
I'll try to update a fair well version maybe few days later.

lpil:
Gunna close this due to inactivity. Please feel free to reopen :)

PR #703, opened 2016-03-20
make default settings for bash more reliable

toogley:
see http://redsymbol.net/articles/unofficial-bash-strict-mode/ for explanation why this might be useful.


lpil:
While I think this snippet is a good addition, I'm not sure it should be the default shebang snippet.

toogley:
Yeah, that arguement makes sense, I'll create a new pull request.

PR #704, opened 2016-03-20
make default settings for bash more reliable

toogley:
see http://redsymbol.net/articles/unofficial-bash-strict-mode/ for explanation
why this might be useful.

improved version of #703

### DIFF
--- a/UltiSnips/sh.snippets
+++ b/UltiSnips/sh.snippets
@@ -33,6 +33,10 @@ snippet !env "#!/usr/bin/env (!env)"
 `!p snip.rv = '#!/usr/bin/env ' + getShell() + "\n\n" `
 endsnippet
 
+snippet sbash "safe bash options"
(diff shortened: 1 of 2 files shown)


lpil:
Thanks :)

toogley:
aah, sorry. i forgot to test it correctly. The Ultisnips  snippet does interpret the \n and \ŧ in IFS=$'\n\t', instead of assigning it to the $IFS variable. Unfortunately, i don't know spontaneously how we can fix this.

lpil:
Oops! We can just remove the ultisnips one, ultisnips will load the snipmate style one quite happily.

toogley:
That, or use that instead:

```
snippet sbash "safe bash options"
#!/usr/bin/env bash
set -euo pipefail
IFS=$'\n\t'
`!p snip.rv ='\n\n' `
endsnippet
```

but i'm not sure what would be the better solution here.

Issue #709, opened 2016-04-01
Latex Greek Symbols


cc7768:
Would people find having shortcuts for greek letters (probably would more or less follow [LatexTools conventions](https://github.com/SublimeText/LaTeXTools/blob/master/LaTeX%20math.sublime-completions)) useful? If so, I can submit a PR with this addition.

honza:
Sounds good to me! :sparkles:

lpil:
Closing due to inactivity. Please feel free to reopen if this needs attention. :)

PR #712, opened 2016-04-06
Add greek letters and a few other math shortcuts.

cc7768:
Greek letters are accessed by a one letter shortcut (except beta where `b` conflicted with a `begin` shortcut in `Ultisnips/tex.snippets`)

Added a few others for things like `\exists`, `\forall`, and the `align*` environment.

Addresses #709 -- Let me know whether this is ok.

### DIFF
--- a/snippets/tex.snippets
+++ b/snippets/tex.snippets
@@ -37,6 +37,10 @@ snippet ali align(ed) environment
 		\label{eq:${2}}
 		${0}
 	\end{align$1}
+snippet ali* align star environment
+	\begin{align*}
+		${0}
+	\end{align*}
 # Gather(ed)
 snippet gat gather(ed) environment
 	\begin{gather${1:ed}}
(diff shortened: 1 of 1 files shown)


lpil:
I think some of these triggers might be a big short. Unsure.

cc7768:
I only chose these triggers because they are what I am used to (and I use them quite often so it is convenient to have 1 key triggers) -- I am not opposed to changing the triggers in the pull request if that is better for others (I would go back to the 1 letter triggers on my own fork after PR is merged).

I did try to make sure they weren't overlapping with other snippet names. If people decide against 1 letter triggers then the next best might be `alpha` triggers `\alpha` -- The `\` is the most annoying part to type.

lverweijen:
For consistency with vim digraphs, I suggest \<letter\>\<star\>.
So a* for \alpha, b* for \beta etc.

cc7768:
This could work -- I like the `<letter><star>` convention.

I can make this change if people are still interested in this pr.

lpil:
Closing due to inactivity. Please feel free to reopen. :)

PR #713, opened 2016-04-06
Fix for R snippets: Close the Else-If snippet

lgeorget:
The Else-If snippet is not closed in file `Ultisnips/r.snippets`. This patch fixes this.


lpil:
Ah, thank you!

PR #714, opened 2016-04-12
Add snippet for casting in CPP

CharlesGueunet:
The  default type for the cast is unsigned because static_cast are often used display a number contained in a char.


lpil:
Thanks!

PR #716, opened 2016-04-19
	snippet `pmdoc`: fix year generate error and make variables more readable

protream:
Sorry, I should test it before.


honza:
No worries :sparkles:

PR #719, opened 2016-04-28
Add snippets for brackets and quotes.

midchildan:
Added snippets like `(` , `{` , `[` , `<` .


lpil:
I don't see the point in these snippets, they don't save any characters.

midchildan:
That seems like a better option. I will close this request.

lpil:
Glad I could help!

PR #721, opened 2016-04-30
Updated PHPUnit, Symfony and Laravel class snippets to be consistent with existing PHP snippets

codeinabox:
Updated PHPUnit, Symfony and Laravel class snippets to be consistent with existing PHP snippets, using a relative path derived namespace and adding `<?php` before class definition

### DIFF
--- a/UltiSnips/php-laravel.snippets
+++ b/UltiSnips/php-laravel.snippets
@@ -36,6 +36,8 @@ endsnippet
 
 #service service provider
 snippet l_ssp "Laravel service provider for service" b
+<?php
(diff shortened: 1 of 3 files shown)


lpil:
Thanks!

codeinabox:
@lpil I noticed the some of the PHP snippets use tabs when the coding standard for PHP these days is spaces. Should I correct this too?

lpil:
The snippet plugin will automatically convert tabs to whatever the user has configured their editor for. We must never use spaces for this reason.

PR #726, opened 2016-05-04
add visual interpolation to tex snippets

psychomario:
Adds the ability to surround visual selection with tex snippets. Shouldn't change default behaviour. See [here](http://vimcasts.org/episodes/ultisnips-visual-placeholder/) for a video.

I've left out a few that do apply but might want python (such as the itemize ones), as I'm not sure how to do them.

Repro:
Visual select, tab, type snippet, tab
visual selection should now be surrounded.


lpil:
Thank you!

PR #728, opened 2016-05-05
PHP and Twig conditional snippets to be consistent with naming convention

codeinabox:


lpil:
This is great, thank you

PR #734, opened 2016-05-15
Add coma at the end of lambda snippets for cpp

CharlesGueunet:
in CPP lambda declarations needs to be coma terminated : [doc](http://en.cppreference.com/w/cpp/language/lambda)


lpil:
Thanks!

PR #735, opened 2016-05-16
Add snippet for `# frozen_string_literal: true`

hwartig:
Please consider inclusion or let me know if you prefer a different snippet trigger. Thanks


lpil:
Ah! How useful! Thanks

PR #742, opened 2016-05-27
Move PHPUnit snipMate snippets into separate file

codeinabox:

### DIFF
--- a/snippets/php-phpunit.snippets
+++ b/snippets/php-phpunit.snippets
@@ -1,0 +1,28 @@
+# phpunit	
+snippet ase
+	$this->assertEquals(${1:expected}, ${2:actual});
+snippet asne
+	$this->assertNotEquals(${1:expected}, ${2:actual});
+snippet asf
+	$this->assertFalse(${1:Something});
+snippet ast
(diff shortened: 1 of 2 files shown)


lpil:
Is PHP unit not ubiquitous in the PHP community??

codeinabox:
It's certainly the most popular, though I assumed you wanted to separate vendor snippets as you have done with all the JavaScript ones. Also this PR removes the superfluous blank lines below each assertion snippet.

lpil:
We did it with the Javascript ones because there are so many JS libraries and frameworks, most of which wanted to use the same triggers. There were too many collisions.

Unless the PHP ones collide or are niche I think they can stay.

Removing the blank lines is a good thing though.

codeinabox:
Okay I'll close this PR and create another one. Would it make sense to move any of the snippets out of UltiSnips/php-phpunit and into snippets/php?

lpil:
It would be better, though it's not a big issue. If you think that's a good use of your time, go for it :)

codeinabox:
Done, created new PR #743

PR #746, opened 2016-06-05
semicolon prepended to iife in js

bounceme:
this is generally agreed upon, even by both sides of the semicolon in js debate,because it will prevent some errors in minifying code.


lpil:
I'm against this modification, but am happy to accept a new snippet that does this in the snipmate format.

PR #751, opened 2016-06-09
Add snippet generating ISO-formatted date

mcepl:
I am a big fan of the ISO date+time format, because it is our only hope
to overcome endless mess which are current zillion non-sensical national
formats.


lpil:
Great! Thank you

PR #755, opened 2016-06-19
Refactored PHP getter and setter snippets to have consistent triggers

codeinabox:


lpil:
Could you remove the ultisnip format ones please? Looks like they just duplicate the shipmate format ones

codeinabox:
The Ultisnips ones are more sophisticated than the Snipmate ones unless the same can be achieved with Snipmate?

lpil:
Ah, I see. OK, looks good

PR #758, opened 2016-06-22
move mocha snippets for javascript folder

Pompeu:


lpil:
This would cause the snippets to be loaded by default, this is not the desired behaviour.

Pompeu:
here dont autoload T_T #758

lpil:
They're not supposed to :)

lpil:
The readme contains information on keeping them load.

lpil:
On making*
Sorry, can't type on a phone so well.

Issue #752, opened 2016-06-11
Self-closing HTML tags


indiesquidge:
Any reason why some HTML snippets are not using self-closing tags for elements, like `input`, `hr`, and `br`, but other elements are using self-closing tags, like `img`?

I'm not sure whether having them is standard with the W3C spec or not at this point, but it seems like things may sometimes still break without them (e.g., `JSX`). Would it be safer to just include them for [void elements](http://w3c.github.io/html/syntax.html#void-elements) in this library? If so, I'd be happy to create a PR to update the snippets.

lpil:
I don't think there's any real reason here. Happy to accept PRs :)

PR #760, opened 2016-07-02
Add self-closing tags to void elements

indiesquidge:
Use self-closing tags on void HTML elements, as discussed in issue #752 

Note that `<keygen>` has been removed completely because it has been [fully deprecated from the Web standards](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen) and should not be used anymore.


lpil:
Cheers :)

PR #762, opened 2016-07-19
By default, use VISUAL selection for HTML tag content

chaucerbao:
Very helpful for marking up existing content.


lpil:
Thank you

PR #765, opened 2016-07-28
Add support for VHDL

robinkjoy:
Adding new language. Feedback is welcome.


robinkjoy:
Wrong folder. Should be in UltiSnips..

PR #769, opened 2016-08-01
Added htmldjango snippets + htmldjango context correction

tweekmonster:
I originally was going to add a couple snippets, but found some problems with the existing ones.  If you prefer that make the fixes a separate PR, let me know.
- Added `include ... with` snippet to htmldjango
- Added `cache` snippet to htmldjango
- Inline block should be allowed anywhere on the line in htmldjango
- htmldjango filter snippets should be allowed anywhere in the line

@carlitux Is there a reason you removed the cursor placeholders in snippets using `${VISUAL}`? https://github.com/honza/vim-snippets/commit/f0a9a1e80b164e05f4b3d1eb6429c01a8f800a6e#diff-5b5193d639ea72b682e4b7c9eafec41bL159

I never use the visual snippet expansion in Ultisnips.  I always assumed the snippets didn't have placeholders.  It's just annoying now that I know the reason ☹️

### DIFF
--- a/UltiSnips/htmldjango.snippets
+++ b/UltiSnips/htmldjango.snippets
@@ -122,6 +122,10 @@ snippet include "" bi
 {% include "${1}" %}
 endsnippet
 
+snippet includew "" bi
+{% include "${1}" ${2:with ${3:var} = ${4:val}} ${5:only} %}
+endsnippet
+
 snippet load "" bi
 {% load ${1} %}
 endsnippet
@@ -156,13 +160,13 @@ endsnippet
 
 snippet with "" bi
 {% with ${1} as ${2} %}
-	${VISUAL}
+	${VISUAL}$3
 {% endwith %}
 endsnippet
 
 snippet verbatim "" bi
 {% verbatim %}
-	${VISUAL}
+	${VISUAL}$1
 {% endverbatim %}
 endsnippet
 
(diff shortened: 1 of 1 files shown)


carlitux:
@tweekmonster hi, I think I was wrong with those changes I was tried the faster edition. Maybe a better option could be set as last jump.

About the fixes that will help I didn't have time to improve the snippets.

tweekmonster:
> Maybe a better option could be set as last jump.

Do you mean `${VISUAL}${2}` instead of `${2:${VISUAL}}`?

carlitux:
${0:${VISUAL}}  I think that should be or ${VISUAL}${0}... what do you think?

tweekmonster:
I didn't know about `$0` because I didn't read far into the docs.  I think I like `${VISUAL}${0}` better.  But, do you think for the inline snippets, it should be something like: `{% static "${VISUAL}$1" %}$0`

carlitux:
well really the $0 next jump is outside the snipped block

tweekmonster:
Hmm, that's weird, when I was testing, `<c-j>` was inserting a new line instead of jumping after the last placeholder.  Maybe I was doing something different.

Anyways, I'll change it so that it uses `${VISUAL}${0}` in the snippets.  Thanks for your help!

tweekmonster:
> Hmm, that's weird, when I was testing, <c-j> was inserting a new line instead of jumping after the last placeholder.

I see the problem now.  `{% static "${VISUAL}$0" %}` causes the cursor's final resting position to be in the parenthesis.  So, an additional `<c-j>` will insert a new line instead of going to the end of the snippet.  It needs to be `{% static "${VISUAL}$1" %}`.  I will update the snippets accordingly.

lpil:
Good to go?

tweekmonster:
@lpil Yup

lpil:
Oops, seems I missed this. Could you resolve the merge conflict please?

lpil:
Closing due to inactivity

PR #770, opened 2016-08-05
fix ada filename mixedcase with - object oriented

JulienPivard:
Derived types use caracter '.' in package name. But in the filename dot
are replaced by '-' .

For exemple :
```ada
package Mother_Package.Daughter_Package is
    -- ...
end Mother_Package.Daughter_Package;
```
is in a file named : Mother_Package-Daughter_Package.ads

So function ada_case(word) change the '-' in filename to dot in the
package name with mixedcase.


lpil:
Thank you

PR #771, opened 2016-08-08
Rename CPP inc snippet to prevent name collision with C inc snippet. Fixes #695.

### LINKED ISSUE #695: Snippet name collision in CPP and C snippets
Both CPP and C snippet files contain a snippet named 'inc'. Since CPP extends C but doesn't seem to override the 'inc' snippet from C, we've to choose which snippet we would like to insert everytime. Is it possible to specify some sort of priority so that CPP snippet is insert by default in CPP files?

dnivra:
I renamed the snippet as incc. Of course, a better name could be chosen if required.

### DIFF
--- a/snippets/cpp.snippets
+++ b/snippets/cpp.snippets
@@ -3,7 +3,7 @@ extends c
 ##
 ## Preprocessor
 # #include <...>
-snippet inc
+snippet incc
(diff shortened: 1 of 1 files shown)


lpil:
Thanks!

PR #773, opened 2016-08-24
Fix incorrect file name pattern - to _

cerobit:
Change - to _  on php_laravel-snippets, php-phpspec.snippets and php-symfony2.snippets, 
UltiSnips not match when -  is used as separator on filenames.


lpil:
This is done to prevent the snippets from being loaded by default.

cerobit:
Ahh Thank you, I understand now,  its cleaner keep optional those optionals and separated for frameworks and specific vendor tools !

lpil:
That's right. Maybe we should have used a clearer naming convention... It's probably too late now

PR #774, opened 2016-08-25
added span, span. and span#

jpgeek:
Not sure if I am missing something, but I couldn't find span, span. or span# in the html snippet, so added one here.

### DIFF
--- a/UltiSnips/html.snippets
+++ b/UltiSnips/html.snippets
@@ -239,6 +239,18 @@ snippet script "XHTML <script>" w
 </script>
 endsnippet
 
+snippet span "<span>" w
+<span> ${0:${VISUAL}} </span>
+endsnippet
+
+snippet span. "<span> with class" w
+<span`!p snip.rv=' class="' if t[1] else ""`${1:name}`!p snip.rv = '"' if t[1] else ""`> ${0:${VISUAL}} </span>
+endsnippet
+
+snippet span# "<span> with ID & class" w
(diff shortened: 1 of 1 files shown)


lpil:
Would it be possible to add these in the shipmate style? That is the preferred format.

jpgeek:
AFAIK, it is not.

Using UltiSnips and removing the UltiSnips directory, it does use the snippets directory with the SnipMate style snippets.  However, when editing eruby this way, it appears to load the snippets/eruby.snippets, but does not include the snippets/html.snippets as it should.

Adding the UltiSnips directory back and editing an eruby file, it loads the UltiSnips/eruby.snippets file and it does include the UltiSnips/html.snippets file.

Not sure if this is a problem with UltiSnips loading of included files in SnipMate style snippets, or if there is a problem with the snippets/eruby.snippets file - I did not see an 'extends' clause there for html, just this comment at the top : "# Includes html.snippets". 

What do you recommend?

lpil:
That seems reasonable. I'm afraid I don't know how to resolve the issue with  the loading of html snippets. :(

jpgeek:
Thanks!

ajzafar:
For what it's worth, SnipMate does support the extends keyword, and [UltiSnips also supports it for SnipMate snippet files](https://github.com/SirVer/ultisnips/blob/4c71935cbaf2264c06bf6f9c7384fa2d7cbe83c8/pythonx/UltiSnips/snippet/source/file/_common.py).

jpgeek:
So it does.  Anyone know why 
snippets/eruby.snippets does not have the extend line for snippets/html.snippets in spite of the comment saying it does?  I am happy to add it in, but it would seem like a huge oversight to not have it there so I suspect someone did that on purpose.

lpil:
Seems like it should be there :)

jpgeek:
aight - will send another pull request.

PR #776, opened 2016-08-31
extended html in snippets/eruby

jpgeek:
Adding extend html to snippets/eruby.snippets.

I tested this using UltiSnips both with and without the UltiSnips directory and it seems to work.  It seems strange that it was missing to begin with, even though there was a comment saying that it *was* included.


lpil:
Thank you.

jpgeek:
Glad to help.

PR #777, opened 2016-08-31
add word boundary flag to ultisnips/javascript `fun` snippet

tragiclifestories:
Motivation: I commonly want to define a callback inline; with the word boundary flag being applied, I fall foul of this old issue https://github.com/SirVer/ultisnips/issues/320.


lpil:
This is great, thanks!

PR #779, opened 2016-09-04
Add webpack snippets

IgorGee:


lpil:
I don't see any requirement for webpack snippets. They are written once per project, so this doesn't really save any time.

The trigger is going to be annoying as they're rather short and general for such niche snippets.

Might I suggest creating your own separate repo for these?

IgorGee:
Hmm. Good point.

PR #786, opened 2016-09-25
Added snippet for GDscript

Aster89:
Added the snippet file with ALL functions of the **GDscript scripting** language of the game engine Godot. The file is **huge** (over 16000 lines), so complex editing commands take much time to be executed.

**This branch contains only edits to this snippet file.** (the last commit rolls back edits I made to other snippets)

### DIFF
--- a/snippets/fortran.snippets
+++ b/snippets/fortran.snippets
@@ -1,3 +1,12 @@
+snippet doc
+	! """"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+	! File: ${2:`vim_snippets#Filename('$1')`}
+	! Author: `g:snips_author`
+	! Email: `g:snips_email`
+	! Github: `g:snips_github`
+	! Description: $1
+	! """"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+	$0
 snippet impl
 	implicit none
 		$0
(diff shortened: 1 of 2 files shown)


lpil:
I feel this collection of snippets is far too large. Why are so many needed?

Aster89:
Why are so many needed?
Honestly, I don't think that a complete collection is needed, in general. I don't think someone could use every single GDscript's function. For what concerns me, I can't remember the name of all the functions I use (even if I don't use so many of them) and, in particular, it's hard for me to remember the arguments of these functions and how they change when applied to different classes. For these reasons I ended up to collect every single function. I just did the work as I think it can be useful to everyone. Then each one can depure it's own fork from unwanted snippets. With Vim it should be an easy task to accomplish.

This doesn't mean that my work is a good work. We can improve it! 😄 

If the question was "Why are so many?" I would answer that the file is so large for two reasons:

1. There are some (a lot, actually) methods of different classes which share the same name,
2. there are methods of a given class that can accept several set of inputs

In the first case, I appended `_name_of_class` to trigger to distinguish (I don't know enough about UltiSnips to use a common trigger for different snippets).
In the second case, I simply commented those snippets. As you can see there are a lot of commented snippets. If you delete those lines, you could end up with a sensibly smaller file.

lpil:
There's no reason to add snippets for every function, especially if the snippet trigger is nearly the same length of the snippet itself. Scanning through it seems that this is all of these snippets.

Instead consider creating snippets that save typing, such as this one:

```javascript
snippet wh
  while (${1}) {
    $0
  }
```

Aster89:
I started assembling this snippet file due to the difficulty I had (and still have) in (a) remembering names of function correctly and (b) the number and type of their arguments. This snippets help me much in this sense, also thanks to (a) UltiSnips' fuzzy logic completion and (b) the fact that I included every single possible way to call a function. If things are just the way you say, then the pull request shoul not be accepted. But, IMHO there is reason to add a snippet for every function, if the problem is that they are too many to remember arguments.

lpil:
I don't think this repo is the right place for these snippets. If you find them useful I recommend creating a repo just for these gdscript documentation style snippets :)

PR #788, opened 2016-10-01
console.warn js snippet

skv-headless:
use it a lot in react native. Feel free to close if you feel that it is unnecessary


lpil:
Thanks!

PR #792, opened 2016-10-02
Missing comment sign at line start

app:
I've got error in latest neovim
UltiSnips.snippet.source.file._base.SnippetSyntaxError: Invalid line 'mi - BEM mix mod' in ~/.config/nvim/plugged/vim-snippets/snippets/javascript/bemjson.snippets:45


lpil:
Thank you. Could you also move these snippets out of the javascript directory please? They shouldn't be in there, my mistake.

PR #796, opened 2016-10-11
ES6 React

mporras:
I would like some feedback before squashing please.


lpil:
Could you move these out of the javascript directory please? Don't want library snippets loaded by default :)

mporras:
@lpil How about now? If it is ok, I'd like to keep doing some changes. Thanks!

lpil:
Thanks!

PR #799, opened 2016-10-11
Add React ES6 UltiSnips

mporras:


lpil:
Shouldn't this be in with the other react snippets?

mporras:
The current ones inside the UltiSnips dir are meant for CoffeeScript.

lpil:
I mean why are they not in the snipmate format file?

mporras:
Because https://github.com/SirVer/ultisnips was not reading them from there. The moment I converted them to UltiSnip format, it started working.

lpil:
Ultisnips reads both formats, but the react snippets are not loaded by default due to the location they are saved to. This is by design- library snippets should be opt-in. :)

mporras:
@lpil Cool! Can we make the ones in this PR, which are in UltiSnips format, to stay?
They look much, much nicer.

lpil:
We can convert any missing snippets into the snipmate format and put them in the other file :)

mporras:
So the snipmate format is the only one that gets to stay?

lpil:
Ultisnips reads both, so unless you need special features it's better to target the snipmate format :)

mporras:
Got it. Since they're not loaded by default, may I ask you how can I get Ultisnips to read them? Thanks!

lpil:
Here you go : https://github.com/honza/vim-snippets#vendor-snippets

mporras:
@lpil Awesome! Thank you :smiley:

PR #802, opened 2016-10-14
Add shebang to Python 3

aldur:
I noticed that there is an analogous snippet in snipmate's `snippets` directory, so I ported it over to UltiSnips.

### DIFF
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -11,6 +11,13 @@ snippet #! "Shebang header for python scripts" b
 $0
 endsnippet
 
+#!3 header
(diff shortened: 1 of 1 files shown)


lpil:
No need for this, ultisnips can use the snipmate one. :)

aldur:
OK, but then what's the point of having other duplicate snippets?

For the record, I had to turn off snipmate support in ultisnips because of [this issue](https://github.com/SirVer/ultisnips/issues/676). If relevant, I can open a different issue for the topic.

lpil:
There is none. Ideally they would be removed, but no one is interested enough to do this dull job.

PR #803, opened 2016-10-18
Short array syntax for php snippets

ernestre:


lpil:
Good call, thanks. Happy to accept more changes like these if there are other snippets using the old format.

Issue #804, opened 2016-10-19
Removing deprecated attributes from HTML table snippet


robingrindrod:
The HTML table snippet in [UltiSnips/html.snippets](https://github.com/honza/vim-snippets/blob/13dbb8a252ebd91b069983f007718a199890258d/UltiSnips/html.snippets#L260) has the _border_, _cellspacing_ and _cellpadding_ attributes. The _cellspacing_ and _cellpadding_ attributes have both been deprecated in the [HTML5 spec](https://www.w3.org/TR/html-markup/table.html#table-constraints). As far as I can see, the _border_ attribute does not seem to deprecated according to the [official spec](https://www.w3.org/TR/html-markup/table.html#table.attrs.border) but the [Mozilla wiki](https://developer.mozilla.org/en/docs/Web/HTML/Element/table#attr-border) seems to think it is. Removing it would bring it in line with pull request #314 and I'd guess the attribute is not widely used.

I'll create a pull request with them all removed as this would be my own personal preference but I'd love to hear other people's opinions.

sedrubal:
Totally agree. Please replace XHTML snippets with HTML5

lpil:
Seems this issue has been fixed.

I think it'd be good to replace XHTML with HTML5, I'd be very happy to merge that PR.

PR #805, opened 2016-10-19
Remove unused attributes from HTML table snippet

robingrindrod:
This removes the border, cellspacing and cellpadding attributes from the
HTML table snippet.

See issue honza/vim-snippets#804


lpil:
Thanks!

PR #808, opened 2016-10-25
Add Proptypes and Default props to React snippets

michaelghinrichs:
This adds proptypes and default props to react snippets in `snipmate`

This was not done for `UltiSnips` as there is no react file. I could port those over in a follow-up if you would like!

:) 

Thank you!


lpil:
Thanks

PR #810, opened 2016-10-26
Prefix rspec describe with "RSpec."

neckhair:
The top most `describe` in an RSpec test should always be namespaced.

So instead of writing
```ruby
describe MyClass do
end
```

you should write

```ruby
RSpec.describe MyClass do
end
```


lpil:
Only for the top, not for the ones below. Because of this I'm -1 to this change. We could have a rdesc snippet for this.

PR #816, opened 2016-11-02
Add C variable printf debug snippets

cirosantilli:
Basically 50% of what I write every day, although it never makes into commits (OK, it does sometimes, but I have to fix remove them later).


lpil:
Thanks! :D

PR #819, opened 2016-11-05
Add snippet for elixir @doc false

craigp:
Just a simple addition for `@doc false` in Elixir, since there's already one for `@moduledoc false`.


lpil:
Nice, thanks!

Issue #806, opened 2016-10-20
[python] `deff` leaves bad placeholder text


sim590:
The snippet is the following:

``` snippets
snippet deff
    def ${1:fname}(${2:`indent('.') ? 'self' : ''`}):
        ${0}
```

Let the following initial file where `^` indicates the cursor position:

``` python
deff
    ^
```

Pressing the expand shortcut, I get the first snippet like so:

``` python
def fname():
    ^
```

Pressing `<ctrl-j>`:

``` python
def fname():
          ^
```

Pressing `<ctrl-j>`:

``` python
def fname(self):

    ^
```

Even though the initial `deff` word was at position 0, the final snippet did insert a `self` word. I guess that this has to do with the instruction `indent('.')` being interpreted at the end of the `ctrl-j` sequence since at the end, it is true that the cursor position (column-wise) is not 0.

sim590:
I must point out that the `def` snippet doesn't have this bug
```snippets
snippet def
	def ${1:fname}(${2:`indent('.') ? 'self' : ''`}):
		"""${3:docstring for $1}"""
		${0}
```

PR #822, opened 2016-11-16
ultisnips.py: add fast def snippet

sim590:
This is addressing the issue of #806. As discussed in this [thread](https://github.com/SirVer/ultisnips/issues/779#issuecomment-260865403), this is most likely an Ultisnips bug. Therefor, the workaround is to provide a ultisnip equivalent snippet so that it is prioritized.


lpil:
Thanks

PR #827, opened 2016-11-28
Indent LaTeX blocks

languitar:
Consistently indent block contents in LaTeX.


lpil:
Thanks! And thanks for the review @Aster89

languitar:
Thanks!

PR #829, opened 2016-12-02
Doxygen snippets were added for C/C++ files

PavloSolntsev:
Doxygen related snippets were added for C/C++ files to provide documentation for functions, enum, struct, and class. Also gpl3 snippet was introduced. If you like it I may add some other too.

### DIFF
--- a/snippets/c.snippets
+++ b/snippets/c.snippets
@@ -125,11 +125,68 @@ snippet do
 ##
 ## Functions
 # function definition
-snippet fun
-	${1:void} ${2:function_name}(${3})
+#snippet fun
+#	${1:void} ${2:function_name}(${3})
+#	{
+#		${4}
+#	}
+# function definition with zero parameters
+snippet fun0
+	/*! \brief ${1:Brief function description here}
+	 *
+	 *  ${2:Detailed description of the function}
+	 * 
+	 * \return ${3:Return parameter description}
+	 */
+	${4:void} ${5:function_name}()
 	{
-		${4}
+		${6}
 	}
+# function definition with one parameter
+snippet fun1
+	/*! \brief ${1:Brief function description here}
+	 *
+	 *  ${2:Detailed description of the function}
+	 * 
+	 * \param $3 ${4:Parameter description}
+	 * \return ${5:Return parameter description}
+	 */
+	${6:void} ${7:function_name}(${8:Type} ${3:Parameter})
+	{
+		${9}
+	}
+
+# function definition with two parameters
+snippet fun2
+	/*! \brief ${1:Brief function description here}
+	 *
+	 *  ${2:Detailed description of the function}
+	 * 
+	 * \param $3 ${4:Parameter description}
+	 * \param $5 ${6:Parameter description}
+	 * \return ${7:Return parameter description}
+	 */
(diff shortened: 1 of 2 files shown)


PavloSolntsev:
It looks like I added some extra spaces. Do I need to clean them and send another pull request? Thanks.

PavloSolntsev:
Just pushed version with correction. Sorry, just learning how to work with people via git.

snippets/c.snippets
```diff
@@ -228,7 +303,33 @@ snippet getopt
...
+# TODO section
+snippet todo
+	/*! TODO: ${1:Todo description here}
+	 *  \todo $1
+	 */
 ## Miscellaneous
 # This is kind of convenient
 snippet .
 	[${1}]
+# GPL v3
```
lpil: No need for this as we already have generic licence snippets

snippets/c.snippets
```diff
@@ -125,11 +125,68 @@ snippet do
 ##
 ## Functions
 # function definition
-snippet fun
-	${1:void} ${2:function_name}(${3})
+#snippet fun
+#	${1:void} ${2:function_name}(${3})
+#	{
+#		${4}
+#	}
```
lpil: Looks like we lost a snippet here

snippets/c.snippets
```diff
@@ -140,21 +197,39 @@ snippet td
...
 	typedef struct ${2:_$1 }{
-		${3:/* data */}
+		m_${3:Data} /*!< ${4:Description} */
 	} ${1:`vim_snippets#Filename('$1_t', 'name')`};
 
 snippet enum
+	/*! \enum $1
+	 * 
+	 *  ${2:Detailed description} 	
+	 */
```
lpil: Can we keep the original doxygenless ones and add new ones for the doced ones? i.e. `enumd`

snippets/c.snippets
PavloSolntsev: Yes, think it make sense to leave it for backward-compatibility. Do you prefer name fun0, fun2 etc or we can add dfun0, dfun1 version for documentation purpose. Thus, we can leave funN versions without documentation and add dfunN version to include documentation. I personally, think that the present documentation block should slightly force people provide some explanation to the instance.

snippets/c.snippets
PavloSolntsev: As I mentioned above, I would propose to add d-prefixed versions with documentation section.

snippets/c.snippets
lpil: Sounds good. :)

PavloSolntsev:
ok, this should be better. dfunX and dmfunX functions were introduced. Protected was used for class instead of private. gpl3 was removed.

PavloSolntsev:
Did you get a chance to review my comment?

lpil:
Sorry to have missed this!

snippets/c.snippets
```diff
@@ -130,6 +130,84 @@ snippet fun
 	{
 		${4}
 	}
+# function definition with zero parameters
+snippet fun0
+	${1:void} ${2:function_name}()
+	{
+		${3}
```
lpil: I'm keen to merge this in as it has been left for a while, this improvement can come in a later PR, thanks. :)

Issue #833, opened 2016-12-18
Wrong output for .tex file with plugin snipmate


lrtfm:
When I insert  `it<TAB>`, I just get a `textit{text}` with out the backslash. It should be `\textit{text}`.
Does we need update the `tex.snippets` file because  snipmate need `\\` to produce a  single backslash?

My `.vimrc` file includes the following lines:
```
let g:snipMate = {}
let g:snipMate.snippet_version = 1
```
Without these two lines, the output is `\textit{${VISUAL:text}}`. It is also a mistake.

How can I fixed these errors?

lpil:
Is this fixed now?

lrtfm:
@lpil [#835](https://github.com/honza/vim-snippets/pull/835) will fixed this.

lpil:
Thank you

PR #835, opened 2016-12-18
Fix output without backslash error for file type .tex (#833)

lrtfm:
1) Add version info.
2) Change single backslash to double backslash.


lpil:
I thought I merged this. Oops.

PR #836, opened 2016-12-19
Add ISC License snippet

mixedCase:
Shorter equivalent of the MIT and simplified BSD licenses.

Default license for packages created with npm.


lpil:
Thanks!

PR #840, opened 2016-12-25
Corrections to reST snippets; Add new snip “eu" and “fnt"

ashfinal:
Minor fixes; Add new reStructuredText snippets: “eu” and “fnt”

eu: Embedded URI
fnt: Footnote or Citation


lpil:
Thank you

Issue #846, opened 2017-01-09
Sass abbreviations expand with ";"


iurevych:
I'm a front-end developer using Sass. When expanding Sass abbreviations they expand just like CSS ones, having ";" after each property. Sass != SCSS as their syntax is different. It is wrong to assume that inside `sass.snippets` you can do `extends css` and add few Sass specific things to it.

Is there an easy fix for it? I'm about to paste all CSS snippets without `;` character into `sass.snippets` manually.

Just wondering if this is just me or everyone else that sees this issue.

In case you need my `.vimrc`: https://gist.github.com/iurevych/927d6c90acd6b05b87378f806ba02402

Here is a screenshot on what's happening as soon as I press `Tab`:
![layouts - vim 2017-01-10 00-22-10](https://cloud.githubusercontent.com/assets/325422/21786071/eb0940c8-d6ca-11e6-85f7-4c72c2f1d6fb.png)

`set ft?` shows "sass", no issues with filetype.

MacVim: 8.0 (120)
Mac OS Sierra 10.12.1

Thanks!

lpil:
I believe the right solution is to remove the extend and then copy and modify the snippets.

PR #847, opened 2017-01-10
Sass snippets based on CSS ones

iurevych:
As @lpil suggested it is better to create new Sass snippets without using `extend css` as the syntax is pretty different between Sass and CSS. I decided to give it a shot. Initial issue created: https://github.com/honza/vim-snippets/issues/846

In the next PRs I can maintain Sass/CSS/SCSS syntax by updating CSS snippets to match up to date W3C specs and look at Emmet for reference.

In the scope of this PR I copied over all CSS snippets and wrapped `url()` into quotes to match the W3C spec.

### DIFF
--- a/snippets/sass.snippets
+++ b/snippets/sass.snippets
@@ -32,9 +32,6 @@ snippet each
 snippet while
 	@while ${1:$i} ${2:>} ${3:0}
 		${0}
-snippet .
-	${1}
-		${0}
 snippet !
 	 !important
 snippet bdi:m+
--- a/snippets/sass.snippets
+++ b/snippets/sass.snippets
@@ -1,14 +1,12 @@
-extends css
-
 snippet $
 	$${1:variable}: ${0:value}
 snippet imp
 	@import '${0}'
 snippet mix
-	@mixin ${1:name}(${2})
+	=${1:name}(${2})
 		${0}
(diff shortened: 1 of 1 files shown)


snippets/sass.snippets
```diff
@@ -34,3 +32,987 @@ snippet each
 snippet while
 	@while ${1:$i} ${2:>} ${3:0}
 		${0}
+snippet .
+	${1}
+		${0}
+snippet !
+	 !important
+snippet bdi:m+
+	-moz-border-image: url('${1}') ${2:0} ${3:0} ${4:0} ${5:0} ${6:stretch} ${0:stretch}
```
lpil: Are these prefix snippets useful?

iurevych: Good point :) in fact they are completely useless as a handful of other ones. Just to keep this PR clear on what it does I decided to keep snippets consistent across CSS and Sass and update both of them at once in the scope of another PR. I plan to review each and every snippet: remove obsolete ones, add missing. But that should be applied to both CSS and Sass snippets together.

lpil: Great. I'd love to see a PR removing some of the uglier CSS snippets :)

lpil:
Thank you!

PR #853, opened 2017-01-19
all.snippets - Use vimscript for datetime snippets

raghur:
On windows, these system date command does not support formatting. vimscript works better


lpil:
How nice, thank you. :)

PR #857, opened 2017-01-24
Sync tex trigger word for itemize and item

kiryph:
for Ultisnips and snipmate:
trigger word for itemize env.: item
trigger word for \item: it (or itd or ]i)


lpil:
Thanks!

Issue #845, opened 2017-01-05
Drop XHTML support and add html5


sedrubal:
I think it's not worth to support XHTML anymore. Instead there should be proper html5 templates.

See [`UltiSnips/html.snippets`](https://github.com/honza/vim-snippets/blob/master/UltiSnips/html.snippets)

related: #804

lpil:
Sounds good to me! :)

jorgesumle:
In `html.snippets`, instead of calling html5 and html5l, we could simply call html and htmll to trigger the base HTML snippets. This would also mean removing the current html (it only does <html>{$0}</html>) snippet, which I think is quite useless in my opinion.

In the same file, we could also remove old DOCTYPES  snippets (XHTML, HTML4...) and only leave the one for HTML 5 DOCTYPE. 

What do you think? I would like to hear opinions before making a pull request trying to impose my changes. If you like this idea, I can make a pull request. I don't if @honza, as HTML Django maintainer, also maintains this or if he should review this first.

lpil:
I don't there is value in renaming snippets without good reason, it will confuse people who are used to typing `html5`. It's only one more character and it's rarely typed.

Much more interested in the smaller more frequently typed ones.

jorgesumle:
Below is the html.snippet I use for HTML. It just supports HTML 5.1. I don't know if it is good to have a snippet to every single HTML tag. Some people probably want just the basic ones. Anyway, tell me what do you think so that me or someone else can make a pull request:

```
priority -50

snippet a
<a href="$1">$2</a>
endsnippet

snippet abbr
<abbr title="$2">$1</abbr>
endsnippet

snippet access
accesskey="$1"
endsnippet

snippet address
<address>$1</address>
endsnippet

snippet article
<article>
	${1:article content}
</article>
endsnippet

snippet aside
<aside>
	${1:aside content}
</aside>
endsnippet

snippet b
<b>$1</b>
endsnippet

snippet blockquote
<blockquote>$1</blockquote>
endsnippet

snippet br
<br>$1</br>
endsnippet

snippet button
<button>$1</button>
endsnippet

snippet caption
<caption>$1</caption>
endsnippet

snippet cite
<cite>$1</cite>
endsnippet

snippet class
class="$1"
endsnippet

snippet code
<code>$1</code>
endsnippet

snippet data
<data value="$2">$1</data>
endsnippet

snippet datalist
<datalist id="$1" name="$2">
	opt$3
</datalist>
endsnippet

[...]

PR #858, opened 2017-01-26
Remove old HTML snippets

jorgesumle:
https://github.com/honza/vim-snippets/issues/845

Remove old stuff of UltiSnips/html.snippets


lpil:
Thanks! This is a great improvement.

PR #861, opened 2017-01-29
Add "nocxx" snippet.

rayvector:
New snippet for the C++ name mangling protection commonly found in the header files of C libraries.


lpil:
Looks like we could remove the ultisnip format and just use the snipmate format possibly

PR #863, opened 2017-01-31
Update snippets for things like if, wh, try to leverage ${VISUAL}

codeinabox:
This is attempt to update many of the existing snippets to use ${VISUAL} and update the docs to encourage people to do this


lpil:
Thanks!

PR #865, opened 2017-02-02
HTML snippets. Remove redundant 'type' attributes

jorgesumle:
#845

>The type attribute allows customization of the type of script represented:
>
> * Omitting the attribute, or setting it to a JavaScript MIME type, means that the script is a classic script, to be interpreted according to the JavaScript Script top-level production. Classic scripts are affected by the charset, async, and defer attributes. **Authors should omit the attribute, instead of redundantly giving a JavaScript MIME type**.

This recommendation is taken from W3C HTML 5.1 Recommendation: https://www.w3.org/TR/html51/semantics-scripting.html#the-script-element. The highlighting was made by me.


lpil:
Thank you. :)

PR #866, opened 2017-02-07
Make python try except snippets py2 and py3 compatible

sedrubal:
```python
try:
   ...
except Exception as err:
    ...
else:
    ...
```

Works for for python3 and python2.


lpil:
Thank you.

PR #867, opened 2017-02-09
Add Elixir @spec snippet

craigp:
Snippet that expands to Elixir type spec notation. I've had this in my
own snippets file for a while and use it regularly.

### DIFF
--- a/snippets/elixir.snippets
+++ b/snippets/elixir.snippets
@@ -63,6 +63,8 @@ snippet wie
 		${4} ->
 			${0}
 	end
+snippet spec
+	@spec ${1:name}(${2:args}) :: ${3:returns}
 snippet df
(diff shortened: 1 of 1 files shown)


lpil:
Cool. We could probably make the snippet shorter, how about `sp`.

Would be good to have ones for `@opaque` and `@type`

craigp:
Good idea, I'll add them to this PR, like the idea of `sp` as well

lpil:
Any progress @craigp ? :)

craigp:
Sorry, I forgot about this, I'll get to this tonight or tomorrow morning

craigp:
Shortened to `sp` and added snippets for `@type`, `@typep`, `@opaque` and `@callback`

PR #868, opened 2017-02-15
Add a TODO comment snippet for PHP.

davidszkiba:
Hi there!

What do you think about this TODO snippet? I came across [Codetags](https://www.python.org/dev/peps/pep-0350/) lately and thought it would be useful to have a snippet for them.

### DIFF
--- a/UltiSnips/all.snippets
+++ b/UltiSnips/all.snippets
@@ -96,4 +96,8 @@ snippet datetime "YYYY-MM-DD hh:mm" w
 `!v strftime("%Y-%m-%d %H:%M")`
 endsnippet
 
+snippet todo "TODO comment" bw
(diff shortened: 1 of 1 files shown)


lpil:
Why limit it to just PHP? Could put it in the general ones with the licences. :)

davidszkiba:
Yeah true, but what about the leading `//` comment? I think not all languages support it, so should I remove the comment or make it optional?

lpil:
Other language agnostic snippets such as box use comments, see how they do it

davidszkiba:
Thanks for the hint about the box comment! Updated the snippet so that it now uses the opening-comment and closing-comment from the `get_comment_format` function. 
And I moved the snippet to `UltiSnips/all.snippets` instead of the licenses file, because I could not import the `get_comment_format` function into `snippets/_.snippets`.

lpil:
Thanks :)

PR #869, opened 2017-02-16
Changed `stn` snippet to use `Self`

saschagrunert:
I think this is more idiomatic Rust since `Self` references to the own structure in Rust.


honza:
Thanks! ✨

Issue #578, opened 2015-05-13
Conforming JavaScript snippets to Standard


mmwtsn:
I've noticed [quite a few projects](https://github.com/feross/standard/blob/master/test/clone.js) adopting [JS Standard code style](https://github.com/feross/standard). There is good support across the more popular editors and its defaults seem fair.

Is there any interest in updating the JavaScript snippets to adhere to Standard?

If so, I'd be happy to open a pull request. If not, is there a recommended way to extend snippets other than installing a fork locally?

lpil:
A near semicolonless Javascript style is definitely not standard. A rather poor name!

If we were to adopt an official JS style we should pick one of the popular ones, rather than an obscure one. Perhaps [Airbnb](https://github.com/airbnb/javascript)?

mmwtsn:
Fair enough.

I think the name comes from the fact that there _isn't_ a standard at the moment. The Node.js community seems to be moving in the direction of semi-colon-less code. Style guides from companies primarily working with languages other than JavaScript on their servers seem to prefer their inclusion.

Both the Airbnb and Google JavaScript style guides are certainly more popular. Maybe Airbnb's then? Their documentation and readability seem better than Google's.

lpil:
The Node.js community has a lot more freedom than the Javascript community at large as they only have to target one interpreter. One of the biggest issues with running JS is that you've no way of knowing what version of what interpreter your code will be run on, so you have to be defensive. i.e. shimming and compiling to older ES# versions. :(

Perhaps. Do we deviate much? I've not checked.

If you want to make some changes, could you make several small PRs rather than one large one? It's much easier to digest that way. :)

mmwtsn:
Absolutely! It's tricky. #468 sort of gets at this problem. Given that most JavaScript files use the same extension it's difficult to know what snippets, exactly, should be sourced or ignored for any given file. As you pointed out client- and server-side JavaScript obviously have very different concerns.

You actually don't deviate much but when linting is set up as part of a build something as trivial as quote style (e.g. single versus double) can prevent the test suite from passing.

Perhaps this has less to do with enforcing a style guide across all the snippets and more to do with extendability. In any event, I forked the repository and dropped the semi-colons to keep my local builds happy :)

lpil:
Aye, #468 annoys me rather a lot.

If you have any clever ideas, please let us know!

mmwtsn:
Understandably. Thanks for talking through this with me!

[...]

PR #870, opened 2017-02-21
Allow configure snippet format for JavaScript

vitalk:
Different JavaScript projects uses different formatting options. The similar issue has been raised a few years ago #578. I trying to find a better way to adopt snippets for them.

This PR uses a python interpolation feature to configure common format options for JavaScript snippets, such as

- __Space before function parenthesis__,

     ```javascript

     function withoutSpace(x) {
         // ...
     }

     var anonymousWithoutSpace = function() {};
     ```

     vs

     ```javascript
     function withSpace (x) {
         // ...
     }

     var anonymousWithSpace = function () {};
     ```

- __Semicolons instead of ASI__,

    ```javascript
    var name = "ESLint";

    var func = function () {
      return 42;
    };
    ```

    vs

    ```javascript
    var name = "ESLint"

    var func = function () {
      return 42
    }
    ```

- __Spacing before and after keywords__,

    ```javascript
    if (foo) {
        //...
    }
    ```

    vs

    ```javascript
    if(foo) {
        //...
    }
    ```

Default formatting is not changed and can be refined via `vimrc`

```viml
let g:ultisnips_javascript = {
      \ 'keyword-spacing': 'always',
      \ 'semi': 'never',
      \ 'space-before-function-paren': 'always',
      \ }
```

or at runtime

```viml
let g:ultisnips_javascript['space-before-function-paren'] = 'never'
```

Since the global snippets works only with UltiSnips format, I add some flow control snippets (`if`, `ife`, `do`) in UltiSnips format as well.

### DIFF
--- a/UltiSnips/javascript.snippets
+++ b/UltiSnips/javascript.snippets
@@ -10,7 +10,7 @@ priority -50
 
 global !p
 from javascript_snippets import (
-	semi, space_before_function_paren
+	semi, space_before_function_paren, keyword_spacing
 )
 endglobal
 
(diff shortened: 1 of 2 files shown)


lpil:
Cool, thank you.

PR #871, opened 2017-02-21
fix notg syntax error in javascript-jasmine

teleivo:
removed extra '.' in the javascript-jasmine.snippets
'notg' snippet which leads to a

SyntaxError: Unexpected token .

        expect(6)..not.toBeGreaterThan(7);
                  ^

if not removed


lpil:
Thanks!

PR #878, opened 2017-02-24
Doxygen and Fortran again

fpnick:
Sorry, I noticed that having a blank line after a parameter documentation line is counter-productive...


lpil:
Thank you.

PR #879, opened 2017-02-28
add javascript-jasmine-arrow.snippets

tompascall:
It is a simple modification of javascript-jasmine.snippets, ie. changing the function calls to arrow functions, in order to be able to use the same coding style in your tests and your other js code.

Here are the diffs between the base snippets and the requested one:
[diff.txt](https://github.com/honza/vim-snippets/files/806935/diff.txt)


lpil:
Thank you.

PR #880, opened 2017-03-01
Added snippet for basic doxygen style cpp fucntion

ParitoshKelkar:
I'm trying to create templates for doxygen style commented c++ code. 
I used python.snippets as a reference for creating the snippet itself. 
(This is my first time contributing to git, please let me know if there is something amiss, or if there is some procedure I didn't follow) 

image

image

image


lpil:
That's great, thank you! Those pictures are really handy too.

PR #883, opened 2017-03-10
Added cs snippets

augustomelo:
There where added the following snippets
- Method
- Constructor
- XML comment parameters


lpil:
Thanks!

PR #884, opened 2017-03-16
Change the options

augustomelo:
This options should be word boundary ('w') because it can only happen after a /// so the option 'b' does not work.


lpil:
Thanks!

PR #886, opened 2017-03-21
Erlang file name is not working any more

johnzeng:
Fixed by this PR. I am using vim-snipmate, not sure if it's caused by wrong snippet driver or not.

### DIFF
--- a/snippets/erlang.snippets
+++ b/snippets/erlang.snippets
@@ -1,14 +1,6 @@
-# module and export all
+# module
 snippet mod
-	-module(${1:`vim_snippets#Filename('', 'my')`}).
-
-	-compile([export_all]).
-
-	start() ->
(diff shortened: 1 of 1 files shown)


snippets/erlang.snippets
```diff
@@ -1,14 +1,6 @@
...
-	-module(${1:`vim_snippets#Filename('', 'my')`}).
-
-	-compile([export_all]).
-
-	start() ->
-	    ${0}
-
-	stop() ->
-	    ok.
+	-module(${1:`vim_snippets#Filename()`}).
```
lpil: Could you make a new snippet for module that exports all? Possibly `modall` ?

Thanks

johnzeng:
Added. @lpil

PR #890, opened 2017-03-26
Create vue.snippets

adriaanzon:
Adds a snippet file for single file Vue components,
consisting of html, javascript and css. 

https://vuejs.org/v2/guide/single-file-components.html

I'm only familiar with UltiSnips so I don't know if other snippet plugins support extending...


lpil:
Thanks :)

PR #892, opened 2017-04-02
Updates C and C++ snippets

logiclogue:
Updates the C/C++ coding styles. As well as separating snippets including documentation and those that don't

Worth taking a look at the `fun3` and `dfun3` commenting fixes for c.snippets. Also worth looking at the `dmfun2` spacing between parameters that was fixed if you don't want to style changes. I noticed there wasn't a space between `catch` and its preceding `{`, for `try` in C++. Don't know if that was intentional.

Hope this helps, otherwise, sorry for wasting your time.

Thanks

### DIFF
--- a/snippets/c.snippets
+++ b/snippets/c.snippets
@@ -42,12 +42,11 @@ snippet #if
 # header include guard
 snippet once
 	#ifndef ${1:`toupper(vim_snippets#Filename('$1_H', 'UNTITLED_H'))`}
-
 	#define $1
 
 	${0}
 
-	#endif /* end of include guard: $1 */
+	#endif /* $1 */
 ##
 ## Control Statements
 # if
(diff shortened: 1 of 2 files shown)


lpil:
Is this a change to make these comply to a particular code style? If so, could you share a link to the documentation for the style please? :)

logiclogue:
Examples from "The C++ Programming Language" use this particular style for member functions: http://www.stroustrup.com/3rd_code.html.

The namespace and class styles are different though.

I've not been coding in C++ for long, so don't know which are the main standards for styles and whether this book is followed for its coding style. If you could point me in the right direction, I'd much appriciate it.

Thanks

lpil:
Perhaps we should pick a style and conform to it.

For C I would suggest the Linux kernel style.

When I search for C++ style the Google style guide is the first item that pops up. Perhaps we should use that one? https://google.github.io/styleguide/cppguide.html

logiclogue:
Yeah that looks like the best one to follow.

Also I've noticed that the C# snippets don't conform to the official conventions either.

I'm happy to update these and make another pull request :)

PR #894, opened 2017-04-09
Latex: Minor Consistency Fixes

etcwilde:
This just makes things more consistent with the other language snippets.

- Makes snippet naming for `\subsubsection*{}` and `\subsubsection{}` consistent between utlisnips and snipmate. 
- Automatically removes square braces from `\usepackage` if no options are supplied, otherwise identical
- Makes un-numbered section, subsection, and subsubsection labels lower case and replaces spaces with underlines in ultisnips.

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -78,6 +78,14 @@ snippet sec "Section" b
 $0
 endsnippet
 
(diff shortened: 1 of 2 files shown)


lpil:
Thank you.

PR #908, opened 2017-05-05
Make TypeScript extend JavaScript for VimL snippets

chaucerbao:
Reflects [UltiSnips/typescript.snippets](https://github.com/honza/vim-snippets/blob/master/UltiSnips/typescript.snippets)


lpil:
Thanks!

PR #912, opened 2017-05-16
Added vue-specific html snippets and vue boilerplate. These two categ…

sanjivj:
…ories were both ported from Sarah Drasner's vue snippets for Sublime Text.


lpil:
Seems there's a merge conflict in the file

lpil:
There seems to be some indentation, some using spaces and some using tabs. Could you swap them all for tabs please?

PR #916, opened 2017-05-20
HTML UltiSnips: HTML 5.1 support

jorgesumle:
Also placed the select snippet after the script tag, as it was not
alphabetically sorted.

This is an attempt to solve https://github.com/honza/vim-snippets/issues/845


lpil:
Nice :)

PR #925, opened 2017-06-01
Go composite literal snippet.

I159:


snippets/go.snippets
```diff
@@ -226,3 +226,13 @@ snippet bench benchmark function
 		}
 	}
 	${0}
+# composite literals
+snippet cl
```
lpil: Why `cl` and not something like `str`?

I159: Composite literal - cl.

lpil:
Thanks :)

PR #926, opened 2017-06-02
added function vim_snippets#FilenameWithExt

darkprof83:
This function may be needed in snippets using the file name. For example, in the license header or doxygen \file parameter.


lpil:
Happy to merge this in if it is used somewhere :)

darkprof83:
I have while only one gpl3 license changed where g:snips_program_description var in my local (for project) .vimrc
[gpl3-snippet.txt](https://github.com/honza/vim-snippets/files/1049164/gpl3-snippet.txt)

PR #928, opened 2017-06-06
fix pdb++

minusf:
there was a typo in pdb++ (pdbpp), it's not pdbbb


minusf:
actually please disregard this.
the actal diff should be removing `pdbpp` entirely as it overrides `pdb` when installed.

PR #929, opened 2017-06-06
pdbpp is a drop-in replacement of pdb, use it importing pdb

minusf:
also:
- it's `pdbpp` not `pdbbb`
- it's not part of `ipython`

:}

### DIFF
--- a/snippets/python.snippets
+++ b/snippets/python.snippets
@@ -142,10 +142,6 @@ snippet ipdb
 snippet iem
 	import IPython
 	IPython.embed()
-# ipython debugger (pdbbb)
-snippet pdbbb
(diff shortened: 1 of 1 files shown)


lpil:
Isn't it this?

https://pypi.python.org/pypi/pdbpp/

minusf:
yes it is.

> Note that the module is called pdb.py so that pdb++ will automatically be used in all places that do import pdb (e.g., py.test --pdb will give you a pdb++ prompt). The old pdb module is still available by doing e.g. import pdb; pdb.pdb.set_trace()

lpil:
Ah, cool, thank you.

PR #930, opened 2017-06-06
Plugin snippet split into author and plugin part.

nuke-ellington:
I've been using these two snippets for quite some time now and thought I'll share them finally.
I understand that the `Plugin` command is an integral part of the snippets system but don't see these changes throwing off people's habits too badly.

### DIFF
--- a/snippets/vim.snippets
+++ b/snippets/vim.snippets
@@ -47,6 +47,6 @@ snippet au augroup ... autocmd block
 		autocmd ${2:BufRead,BufNewFile} ${3:*.ext,*.ext3|<buffer[=N]>} ${0}
 	augroup end
 snippet bun Vundle.vim Plugin definition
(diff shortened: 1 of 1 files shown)


lpil:
This costs more keystrokes with the default Ultisnips config as the key binding for moving to the next field is more typing than the `/`. Also makes it harder to copy/paste in a repo name from GitHub as only the first field is overwritten.

nuke-ellington:
Haven't really thought about the copy/paste case since I usually find it easier to type it all out, but good point.

lpil:
You could add it as a new snippet if you like :)

nuke-ellington:
Since this appears to be personal pref I put it into `.vim/after/snippets` locally.

[https://github.com/nuke-ellington/after-snippets](url)

PR #934, opened 2017-06-09
Blade UltiSnips: make echo statements trigger also just after a non-whitespace c…

jorgesumle:
…haracter

It is very inconvenient if you want to trigger an echo snippet inside an HTML tag (e.g. `<h1>{{  }}</h1>`). This commit solves that


lpil:
Thanks!

PR #941, opened 2017-06-22
Blade UltiSnips: add snippet for @lang directive

jorgesumle:
documented https://laravel.com/docs/5.4/localization#retrieving-translation-strings

### DIFF
--- a/UltiSnips/blade.snippets
+++ b/UltiSnips/blade.snippets
@@ -76,6 +76,10 @@ snippet inject "@inject directive"
 @inject('$1', '$2')
 endsnippet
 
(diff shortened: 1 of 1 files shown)


lpil:
The other snippets format is more widely supported, would you mind removing this one and adding it in the other format as it does not use any of the special ultisnips features?

jorgesumle:
Sorry, what do you mean? Can you point me to an example of the more supported format?

lpil:
Oh, I see we only have blade in the ultisnips format currently. That's a shame. I'll merge this in and we can fix that at a later date. Thank you!

PR #942, opened 2017-06-23
Make sh's tmpfile easier to use

idbrii:
mktemp requires some X in the filename and the current snippet does not include them. Add some X.

Also adds code to supply a suffix which requires a special argument.


lpil:
Thank you

PR #948, opened 2017-07-01
Add missing semicolons after assert macros in rust.snippets

ynakao:
Add semicolons at the end of `assert!` and `assert_eq!`.


honza:
Thanks!

PR #950, opened 2017-07-03
GO: Json marshalable object

I159:
To avoid infinity recursion when MarshalJSON method is defined required
to declare an alias type to wrap an original struct.


lpil:
Thank you

Issue #952, opened 2017-07-13
javascript.es6.react.snippets not working..


fdelacruz:
I tried doing as per the documentation:

- :UltiSnipsAddFiletype javascript.es6.react

- used .vim/ftplugin/javascript.vim

- autocmd FileType javascript UltiSnipsAddFiletypes javascript.es6.react

..to no avail.

On the other, hand javascript-mocha which is in the same path is working fine which means it is NOT the snippet engine.

fdelacruz:
Any idea what so ever, as to what I'm doing wrong??

fdelacruz:
So not one of the four contributors of the thing @michaelghinrichs, @mauroporrasp, @bannier, @codeinabox has a word to say about this issue.

codeinabox:
To confirm are you using Ultisnips or a different snippet engine?

fdelacruz:
Hi @codeinabox, I'm using Ultisnips.

codeinabox:
@fdelacruz first off sorry you haven't received an answer. This project is run by volunteers, if you want answer from a contributor it's good to start with please rather than just mentioning them.

I've had a look into this and I am wondering if the issue is it uses dots instead of dashes. @lpil I was under the impression that dashes are used for vendor snippets and dots for multi-syntax snippets? I could make a PR to change it if that's the problem.

lpil:
Hi @fdelacruz, please be civil and polite when asking people to donate their time to help you for free.

I feel does should work but I know little about the details of this ultisnips feature. I'm on a train now so can't investigate right now I'm afraid

bannier:
@fdelacruz doesn't work anymore for me too, nor mocha snippets. looks like javascript related snippets won't load outside of `javascript` directory

bannier:
moving `javascript.es6.react.snippets` file to `javascript` directory solves the issue for me

lpil:
If we put it there it is loaded by default, which may not be the behaviour we want. Do we want to make ES6 the default? If so we should probably move some or all the snippets into the regular javascript file.

fdelacruz:
Apologies for the rudeness earlier.
That said, thanks for taking a look a this.
As for javascript-mocha.snippets, mine are working fine.

fdelacruz:
@codeinabox using dashes instead of dots seems to fix the problem.

code-hunger:
[...]

PR #955, opened 2017-07-24
fix(#952): javascript.es6.react.snippets not working

bannier:
moves javascript.es6.react.snippets to javascript directory as it doesn't seems to work without it


lpil:
This enabled the ES6 snippets by default, is this what we want? Is ES6 to be considered the default now?

bannier:
Hi @lpil,
You're right : I didn't realized what it implied.

Personally I think ES6 should become the default as it is almost entirely supported by [many browsers](https://kangax.github.io/compat-table/es6/) and [node](http://node.green/).
What's the next practical steps from here : someone in the core maintainers decides ? wait for some form of votes ?

lpil:
We could open a discussion to talk about ES6 snippets, but this change can't be merged in as it also makes React snippets default. Most Javascript users will not be React users.

bannier:
OK this is a valid concern. Closing the PR then.

lpil:
Thank you :)

PR #959, opened 2017-08-20
added `pirv` and `prot` to UltiSnips snippet

bishwahang:
* Added `priv` and `prot` snippet in UltiSnips. Without this, it triggers the `snippets/ruby.snippets`; but this will add indentation/extra whitespace, in the empty line below the `private` keyword and before cursor position.

  Actual:
  ```ruby
  private
  _ _
  |
  ```
   Expected:
  ```ruby
  private

  |
  ```
* Also, indented cursor one level deep after `when` in `case` statement snippet.

### DIFF
--- a/UltiSnips/ruby.snippets
+++ b/UltiSnips/ruby.snippets
@@ -302,7 +302,7 @@ endsnippet
 snippet "\b(case|sw(itch)?)" "case <variable> when <expression> ... end" r
 case ${1:variable}
 when ${2:expression}
-$0
+  $0
 end
 endsnippet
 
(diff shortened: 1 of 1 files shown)


lpil:
Is there a problem with the snipmate format snippet? We should avoid adding ultisnips snippets unless the snippet requires special ultisnips features.

bishwahang:
Snipmate format works okay, but for that you need to remove `SirVer/UltiSnips` and install `garbas/vim-snipmate`. If you have `SirVer/UltiSnips` installed, then there is extra white space on the empty line before the cursor.

Looks like this issue of extra white space was raised here, https://github.com/SirVer/ultisnips/issues/391, and solve with this PR, https://github.com/SirVer/ultisnips/pull/434/files.

lpil:
Ah, we're using `m`, cool :)

Thanks!

PR #962, opened 2017-09-01
Add pygal chart for rst

ashfinal:
pygal chart for Nikola — Static Site Generator

### DIFF
--- a/UltiSnips/rst.snippets
+++ b/UltiSnips/rst.snippets
@@ -302,7 +302,7 @@ snippet fnt "Footnote or Citation" i
 endsnippet
 
 # Only for Nikola — Static Site Generator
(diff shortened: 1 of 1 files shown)


lpil:
This snippet seems like it is for a specific library rather than for the language as a whole. Should it be included by default? If so I think we should use a longer trigger.

ashfinal:
It's one directive of reStructuredText Extensions which were introduced by Nikola. https://getnikola.com/handbook.html#restructuredtext-extensions I've found it's really handy for producing SVG charts while blogging, and there is no other ways to produce charts with reStructuredText, as far as I know. 

But you are right. Since it's not a standard directive of reStructuredText, using a longer trigger seems better.

PR #963, opened 2017-09-01
Add snippet for given when then for python tests.

grangerp:


lpil:
Could you move this over to the snipmate format please? That format is supported by all snippet plugins :)

grangerp:
I have move the snippet to the snipmate format.

lpil:
Thanks!

PR #967, opened 2017-09-06
Removing unnecessary empty line in doc directives

craigp:


lpil:
These are part of the recommended Elixir compiler used by the core team, so I'm in two minds about removing them. I suppose most projects don't conform to this style...

craigp:
Interesting, I hadn't noticed that before - but looking through the elixir-lang source I see that they do seem to do that.

lpil:
This can be your call. What do you think we should d here?

craigp:
It's a tough one, we try to emulate what the core team does, but personally I don't like it in my code. I guess if people want the extra space they can just hit Enter..

PR #968, opened 2017-09-06
basic genserver structure

craigp:


snippets/elixir.snippets
```diff
@@ -197,3 +197,15 @@ snippet >i pipe to inspect
 	|> IO.inspect
 snippet >il pipe to inspect with label
 	|> IO.inspect(label: "${1:label}")
+snippet gs basic genserver structure
```
lpil: This is going to be used infrequently, so could we give it a longer trigger please? :)

snippets/elixir.snippets
```diff
@@ -197,3 +197,15 @@ snippet >i pipe to inspect
...
+	use GenServer
+
+	@doc false
+	def start_link do
+		GenServer.start_link(__MODULE__, ${1:Map.new}, name: __MODULE__)
```
lpil: We should not default to naming the process. Naming process by default removes the ability to test them concurrently.

snippets/elixir.snippets
craigp: Sure - in my personal snippets it was `genserver`, so I'll change it to that

snippets/elixir.snippets
craigp: True, I just don't often create unnamed processes - I'll change it.

PR #975, opened 2017-09-20
Python comprehensions

yangmillstheory:
Reopening #973 to merge from feature branch. Also fixed a `SnipMateSyntaxError` for `snippets/python.snippets`.

### DIFF
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -709,3 +709,4 @@ snippet pmdoc "pocoo style module doc string" b
 $0
 endsnippet
 
+# vim:ft=snippets:
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -709,4 +709,16 @@ snippet pmdoc "pocoo style module doc string" b
 $0
 endsnippet
 
+##################
+# Comprehensions #
+##################
+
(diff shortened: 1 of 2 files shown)


UltiSnips/python.snippets
```diff
@@ -709,4 +709,16 @@ snippet pmdoc "pocoo style module doc string" b
...
+# Comprehensions #
+##################
+
+snippet lcp "list comprehension"
+[$1 for $2 in ${3:${VISUAL}}]$0
+endsnippet
+
+snippet dcp "dict comprehension"
+{$1: $2 for $3 in ${4:${VISUAL}}}$0
+endsnippet
```
lpil: Remove these please, no reason to duplicate the snippets if they don't use any Ultisnips specific features :)

yangmillstheory: Sorry, I'm not sure what you mean, what should I remove? I'm new to developing against UltiSnips, just a casual user.

lpil: No problem! All plugins can read the snipmate format, so there's no need to duplicate snippets in both formats. The two you've added in the ultisnips format can be safely removed :)

yangmillstheory: Thanks, mind blown 👍 

mindblown

PR #982, opened 2017-09-25
p to pipe operator snippet for elixir

josephan:
I noticed there was a missing snippet simply for `|>` for elixir which is used frequently.
This has prevented me from repositioning my hand every time I wanted to input a pipe operator.

### DIFF
--- a/snippets/elixir.snippets
+++ b/snippets/elixir.snippets
@@ -184,7 +184,7 @@ snippet beh
 	@behaviour ${1:Mix.Task}
 	${0}
 snippet p
-	|>
+	|> ${0}
 snippet >e pipe to each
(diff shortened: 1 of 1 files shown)


snippets/elixir.snippets
```diff
@@ -183,6 +183,8 @@ snippet qu
 snippet beh
 	@behaviour ${1:Mix.Task}
 	${0}
+snippet p
+	|>
```
lpil: How about `|> ${0}` so there's a space after as well? :)

lpil:
Nice idea! I've left a comment above

snippets/elixir.snippets
josephan: You just made that snippet 2x better! Thanks!

PR #983, opened 2017-09-26
Adding set comprehension

yangmillstheory:
Forgot that this was possible, so this is more of an addendum to #975.

```
🤔 ~/c/p/vim-snippets (2a84212)|py-comprehensions✓
± [i]: python -c "print {j for j in xrange(10)}"
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```


yangmillstheory:
Can we squash merge this in for the cleaner commit history?

PR #996, opened 2017-10-22
Consolidated all javascript snippet files into the same folder

cjbassi:
The javascript files in the `snippets` folder were kinda scattered, but now they're all in the same folder.


cjbassi:
It turns out there were 2 react javascript snippet files so I just combined those too.

cjbassi:
Let me know if any of the snippet names should be changed.

lpil:
The Javascript snippets are kept in those places so they are not loaded by default and so that people can explicitly load them.

cjbassi:
Okay that makes sense. It might be good to mention that not all snippets are intended to be loaded by default tho, because I thought it was originally a bug instead of the intent.

Also, there seems to be some discrepancy between the `snippets` folder and the `UltiSnips` folder about what gets loaded by default. For example, `javascript-node.snippets` is loaded in `snippets` but not in `UltiSnips`.

lpil:
It is documented in the readme :)

cjbassi:
Lol woops. Thanks for pointing that point.

lpil:
Thanks :)

PR #998, opened 2017-10-23
Add a javascript-ui5.snippet file

Trulsaa:
Add a openui5 snippet file that works with Ultisnips, in place of the one recently moved to the snipmate snippets folder.


lpil:
There's no need to duplicate these snippets as Ultisnips can read the snipmate format.

PR #1000, opened 2017-11-02
remove extra indent from UltiSnips html.snippets

mwcz:
In the UltiSnips html.snippets, the `html` and `htmll` snippets begin with a tab on each line, which leads to the entire HTML document having (to my eyes) an extra level of indentation.  There may be a reason for this that I'm not seeing, but I prefer to not have the tab and thought I'd send in a PR for this change.  Thanks for the awesome snippet engine!


honza:
Thanks! :sparkles:

PR #1001, opened 2017-11-08
robot: fix Set Variable keyword

mMontu:
The syntax of `Set Variable` is slightly different from `Set Global Variable`, `Set Test Variable`, and `Set Suite Variable`.


lpil:
Thanks

PR #1002, opened 2017-11-13
Remove annoying spaces within <span>

qstrahl:
For some reason, the `span` snippets in `UltiSnips/html.snippets` contain these unnecessary spaces around the inner content. This is inconsistent with all the other inline tags - and also with the `span` snippets in `snippets/html.snippets`. This patch removes those spaces.


lpil:
Much better, thanks

PR #1012, opened 2018-01-04
Add cpp anonymous namespace snippet

blukat29:
Adds 'ans' keyword for c++ anonymous namespace.

### DIFF
--- a/snippets/cpp.snippets
+++ b/snippets/cpp.snippets
@@ -138,6 +138,11 @@ snippet ns
 	namespace ${1:`vim_snippets#Filename('', 'my')`} {
 		${0}
 	} /* namespace $1 */
+# anonymous namespace
+snippet ans
+	namespace {
+		${0}
+	} /* anonymous namespace */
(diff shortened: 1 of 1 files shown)


snippets/cpp.snippets
```diff
@@ -138,6 +138,11 @@ snippet ns
 	namespace ${1:`vim_snippets#Filename('', 'my')`} {
 		${0}
 	} /* namespace $1 */
+# anonymous namespace
+snippet ans
+	namespace {
+		${0}
+	} /* anonymous namespace */
```
lpil: Is this commented needed?

blukat29: Which comment are you referring to? `# anonymous namespace` is not needed. But `/* anonymous namespace */` should be there at least for consistency with the snippet 'ns'.

lpil: Could we remove it from both please?

lpil:
Thanks!

PR #1014, opened 2018-01-06
Update css.snippets

abra:
It will be convenient to have a default snippet for "background-image"


abra:
Fair enough. Somehow it was left without my attention. You can reject this pull-request! :)

lpil:
OK, if you're sure! :)

PR #1016, opened 2018-01-09
Rust: Improve pri, pln, and fmt

phansch:
Using the snippets with format args pretty much always means that the
user wants to use the format args using `{}` or `{:?}`, so it makes sense
to include at least the `{}` in the snippet.

See: https://doc.rust-lang.org/std/fmt/

### DIFF
--- a/snippets/rust.snippets
+++ b/snippets/rust.snippets
@@ -42,13 +42,13 @@ snippet lettm "let mut variable declaration with explicit type annotation"
 snippet pri "print!"
 	print!("${1}");
 snippet pri, "print! with format param"
-	print!("{${1}}", ${2});
+	print!("${1}{${2}}", ${3});
 snippet pln "println!"
 	println!("${1}");
 snippet pln, "println! with format param"
-	println!("{${1}}", ${2});
+	println!("${1}{${2}}", ${3});
 snippet fmt "format!"
-	format!("{${1}}", ${2});
+	format!("${1}{${2}}", ${3});
 
 # Modules
 snippet ec "extern crate"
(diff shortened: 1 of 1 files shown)


lpil:
Surely you'd want the cursor outside the `{}`?

phansch:
Hmm, you raise a good point. I'd want it inside the `{}` if I know that it will be a `{:?}`, instead of the simple `{}`. I'd also want it inside the `{}` if the format args are named like
```rust
println!("{a}", a=1);
```

On the other hand, I would want it outside the `{}` if I want to have same text before or after the interpolated argument:

```rust
println!("some_var: {}", some_var);
```

How about adding an additional cursor jump in front of the `{}`?

phansch:
I added the additional jump in front of the `{}` as well now. 

[rustbyexample/print](https://rustbyexample.com/hello/print.html) has a lot of examples, too. Essentially I would like to be able to put text in front to label the output and also be able to how the variables should be *formatted*.

```rust
// Simple print using 'display' trait (like a single number or string)
println!("item: {}", item_that_implements_display);

// Debug print using 'debug' trait (like arrays or any other more complex type)
println!("item: {:?}", item_that_implements_debug);

// With named args and 'display' trait
println!("item: {abc}", abc=item_that_implements_display);

// With named args and 'debug' trait
println!("item: {abc:?}", abc=item_that_implements_debug);

// Pretty printing
println!("item: {:#?}", item_that_implements_debug);

// This can also be used with println/print/etc
format!("{:04}", 42); // => "0042" with leading zero

// Right-padding a string:
println!("Hello {1:0$}!", 5, "x"); // => "Hello x    !"
```

lpil:
Thanks!

Issue #1020, opened 2018-01-23
MochaJS shouldn't use fat arrow syntax


hovissimo:
https://github.com/honza/vim-snippets/blob/1fd6d97e64932191b9c3f3095497bef8fd191f65/snippets/javascript-mocha.snippets#L1-L8

Fat arrows are discourage in Mocha JS usage, see https://github.com/mochajs/mocha/issues/2018

lpil:
Closing due to inactivity. Please reopen if this is a current problem.

PR #1021, opened 2018-01-23
#1020 Update mochajs snippets to use old functions

hovissimo:
See https://github.com/honza/vim-snippets/issues/1020
See https://github.com/mochajs/mocha/issues/2018

### DIFF
--- a/snippets/javascript-mocha.snippets
+++ b/snippets/javascript-mocha.snippets
@@ -1,29 +1,29 @@
-snippet des "describe('thing', () => { ... })" b
-	describe('${1:}', () => {
+snippet des "describe('thing', function() { ... })" b
+	describe('${1:}', function() {
 		${0:${VISUAL}}
 	});
-snippet it "it('should do', () => { ... })" b
-	it('${1:}', () => {
+snippet it "it('should do', function() { ... })" b
+	it('${1:}', function() {
 		${0:${VISUAL}}
 	});
-snippet xit "xit('should do', () => { ... })" b
-	xit('${1:}', () => {
+snippet xit "xit('should do', function() { ... })" b
+	xit('${1:}', function() {
 		${0:${VISUAL}}
 	});
-snippet bef "before(() => { ... })" b
-	before(() => {
+snippet bef "before(function() { ... })" b
+	before(function() {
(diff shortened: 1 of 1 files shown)


lpil:
Hi @hovissimo. Why do you want to make this change? The only versions of node that do not support this are no longer supported

- http://node.green/#ES2015-functions-arrow-functions
- https://github.com/nodejs/Release

hovissimo:
@lpil

See https://github.com/mochajs/mocha/issues/2018 from the PR description.

Fat arrows are specifically discouraged for Mocha JS because Mocha JS uses 'this' for test context. The lexical binding of the fat arrows syntax prevents that.

lpil:
Thanks for the link, very helpful.

There's reason to use the long-form function syntax, but in the wild I've only seen the arrow syntax used, so it seems there's a split here. I think it'd be good to make this configurable by the user, they can set a value indicating which to use.

I'd like to default to the arrow to match existing behaviour.

Pompeu:
> @lpil
> 
> See [mochajs/mocha#2018](https://github.com/mochajs/mocha/issues/2018) from the PR description.
> 
> Fat arrows are specifically discouraged for Mocha JS because Mocha JS uses 'this' for test context. The lexical binding of the fat arrows syntax prevents that.

You can use config file of mocha, and dont need use "this".

hovissimo:
@Pompeu I don't see how any options in the config file change the lexical binding of test blocks. Can you point to exactly what you mean?

I maintain that if this project is going to maintain snippets for a specific testing library it should conform to that library's most general recommendations. Further, I find fat arrows for the sake of consistency to be a weak argument, because fat arrows are in no way replacing the old function declaration. They're a new and different kind of function declaration per the ES spec.  One does one thing (methods), the other does another (subroutines/procedures) and the right tool should be used for the job.  In this case these are methods and NOT subroutines and the authors of mochajs picked methods in particular and documented their decision (and included a specific warning against arrow functions!).

 That said, it's your project and I've already forked it for my use so /shrug.

lpil:
Closing due to inactivity. Please re-open if you'd like to continue. Thank you! :)

PR #1031, opened 2018-03-12
classheader and defheader using puppet strings

jvidal-sysadmin:
These snippets are a bit old and I will try to adapt them to the new style guides. My first big change will be how to document "classes" and "define" using "Puppet strings" (https://github.com/puppetlabs/puppet-strings).

On the other hand I would like to receive suggestions from the community and include them so that this collection of snippets has much more quality.

I am working on including all the most basic snnipets: https://puppet.com/docs/puppet/latest/type.html


lpil:
Thanks

PR #1035, opened 2018-03-19
Allow quotes to be configured for ruby snippets

juanibiapina:
This is a prototype for using a configuration for seleting the type of quotes
for ruby files. I would like to get some feedback before completing the changes
in all ruby snippets.


lpil:
Seems sensible to me. I'm not sure we have a best place for that

juanibiapina:
I'm not sure I'll follow up on this. It turns out it's so much work that slowly migration the snippets that I use to my own repo is faster.

Anyone wants to pick it up or should I close it?

PR #1038, opened 2018-03-29
JavaScript snippets - React Component and getElementsByClassName

junkdeck:
modified the 'get' snippet for UltiSnip to accept "C" as an option for `getElementsByClassName` and added 'irc' to directly import the Component class from React. hope you find them useful!

also renamed the anonymous function snippets to avoid conflicts with the ES6 arrow function snippets.

### DIFF
--- a/UltiSnips/javascript.snippets
+++ b/UltiSnips/javascript.snippets
@@ -68,7 +68,7 @@ ${1:var }${2:function_name} = function $2`!p snip.rv = space_before_function_par
 }`!p snip.rv = semi(snip)`
 endsnippet
 
-snippet af "Anonymous Function" i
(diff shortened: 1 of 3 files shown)


UltiSnips/javascript.snippets
```diff
@@ -68,7 +68,7 @@ ${1:var }${2:function_name} = function $2`!p snip.rv = space_before_function_par
 }`!p snip.rv = semi(snip)`
 endsnippet
 
-snippet af "Anonymous Function" i
+snippet anf "Anonymous Function" i
```
lpil: Why have we changed this?

junkdeck: it conflicted with `af` for arrow functions.

PR #1042, opened 2018-04-17
Add a go snippet for HTTP handler methods

crdueck:
Following the pattern of the existing `func`/`funch` snippets, this
commit adds a `fumh` snippet to go along with `fum`.

Note: `UltiSnips/go.snippet` provides the `func` & `funch` snippets, while `snippets/go.snippet` provides the `fum` snippet that this `fumh` pairs with. I wasn't quite sure where this should live, but decided on `snippets/go.snippet`. Let me know if it should live in `UltiSnips/` instead.

### DIFF
--- a/snippets/go.snippets
+++ b/snippets/go.snippets
@@ -133,6 +133,11 @@ snippet fum
 		${6}
 	}
 	${0}
+# http handler function on reciever
+snippet fumh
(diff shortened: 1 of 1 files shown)


lpil:
Snippets is the right dir, thanks!

PR #1051, opened 2018-05-16
add VISUAL placeholder; remove option "b"

heiderich:


lpil:
Why are we removing the b option? :)

heiderich:
I think in a typical situation where the VISUAL placeholder would be used, the cursor might not be at the beginning of the line. Therefore I removed the "b" option.

lpil:
What about when visual is not being used?

PR #1052, opened 2018-05-19
Add error handling snippet for Go using errors.Wrap()

pmalek:


lpil:
Could you add these to the snipmate format rather than the ultisnips format? Then it can be used by all editors

pmalek:
@lpil I can do it no problem but I thought that ultisnips is the preferred format now since snipmate is an almost dead project now

lpil:
The snipmate format is the most widely supported and is our default here.

As for the snipmate plugin, it's not updated bit that's largely because it doesn't need to be, it works :)

pmalek:
@lpil done. Please re-check.

PR #1062, opened 2018-07-15
markdown text formatting

heiderich:
*foo*
**foo**
***foo***

### DIFF
--- a/UltiSnips/markdown.snippets
+++ b/UltiSnips/markdown.snippets
@@ -50,6 +50,23 @@ snippet spar "Paragraph" b
 $0
 endsnippet
 
+###################
+# Text formatting #
+###################
+
+snippet * "cursive"
+*${1:${VISUAL}}*$0
+endsnippet
+
+snippet ** "bold"
+**${1:${VISUAL}}**$0
+endsnippet
+
+snippet *** "bold"
+***${1:${VISUAL}}***$0
+endsnippet
+
+
 ################
 # Common stuff #
 ################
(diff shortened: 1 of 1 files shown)


UltiSnips/markdown.snippets
```diff
@@ -50,6 +50,23 @@ snippet spar "Paragraph" b
...
+
+snippet * "cursive"
+*${1:${VISUAL}}*$0
+endsnippet
+
+snippet ** "bold"
+**${1:${VISUAL}}**$0
+endsnippet
+
+snippet *** "bold"
```
lpil: Is this text right?

UltiSnips/markdown.snippets
```diff
@@ -50,6 +50,23 @@ snippet spar "Paragraph" b
 $0
 endsnippet
 
+###################
+# Text formatting #
+###################
+
+snippet * "cursive"
```
lpil: Would a letter based trigger be better? These are normally easier to type.

Would "italics" be a better description? It doesn't output cursive text.

heiderich: I like the asterisk(s) as a trigger. This way the user does not need to remember another trigger.

I agree. "italics" would be a better description. I corrected it in b8f54d395e2f0b20b0ae12e3e21dddad2a0b3fda.

It seems that the concrete formatting is indeed not specified in the commonmark specification. However the descriptions I used seem to describe well the concrete "effect" of the asterisks in common implementations (at least those I used).

UltiSnips/markdown.snippets
heiderich: No, this was a copy & paste error. I corrected it in b8f54d395e2f0b20b0ae12e3e21dddad2a0b3fda.

PR #1088, opened 2018-12-08
Adding kubernetes helm snippets ..

vinodhalaharvi:
Kubernetes has several more objects and I will try to add them in the future. But this should cover two most used kubernetes objects -  Deployments and services.


lpil:
Could you move these out of the `yaml` directory please? Kubernetes snippets should not be loaded by default for all yml files, instead they should be opt-in by the user https://github.com/honza/vim-snippets#vendor-snippets

vinodhalaharvi:
done, thanks

lpil:
Thank you

PR #1089, opened 2018-12-12
Update ember-snippets to Ember v3

dcyriller:
Ember framework has changed a lot since July 2015! This PR aims at updating UtliSnips snippets to v3.

### DIFF
--- a/UltiSnips/javascript-ember.snippets
+++ b/UltiSnips/javascript-ember.snippets
@@ -4,57 +4,45 @@
 
 priority -50
 
-# Application
-snippet eapp "App.Name = Ember.Application.create({});"
-${1:App.Name} = Ember.Application.create({});
+# imports
+snippet eiapp "import Application"
+import Application from "@ember/application"
 endsnippet
 
-# Models
-snippet emod "App.ModelName = Ember.Model.extend({...});"
-${1:model_name} = Ember.Model.extend({
-	${0://Properties here...}
-});
+snippet eicon "import Controller"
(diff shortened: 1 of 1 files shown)


lpil:
It seems we've replaced snippets for using these entities with snippets for importing them. I'm not an ember user but this seems like we've lost some functionality here.

Could you either put them back or update them for the new conventions while keeping the triggers the same please? Thanks :)

dcyriller:
Thank you for your quick answer :)

I'm thinking about it and I realize the way we create objects in Ember have changed dramatically since the snippets were setup... 

We would now use a cli generator to create the entities. This is why I simply put the imports at first. But with your comment, I realize that most of the imports would not be usefull. 

I did this PR because I needed the "computed properties" snippets. And, actually, the "computed property" snippets would probably be enough. I'll update the PR to match the triggers for the computed properties and remove the unneeded imports.

I've asked some fellow Ember users to review the snippets.

lpil:
Even if generators are now popular we don't want to remove functionality from this snippet repo. Some people may be using them and there is no cost to keeping them here :)

dcyriller:
Ok, I understand. I've added back the removed snippets.

I kept the view one removed (since this class is not exposed anymore), and added a Component one.

Feel free to squash the commits if this sounds acceptable to you :)

lpil:
Thank you for your contribution and your patience with my feedback!

dcyriller:
Thank you for your review 👍

PR #1090, opened 2018-12-14
Use same trigger for all javascript functions

idbrii:
I assume there were different triggers for technical limitations, but [SirVer explained here how to have a "not at beginning of line trigger"](https://github.com/SirVer/ultisnips/issues/400).

For using like this:
    fun<Tab>
    =fun<Tab>
    setTimeout(fun<Tab>

None invoke the snippet
selector menu.

The anonymous function snippet also lets you jump to the end to close
scope before defining the function.


UltiSnips/javascript.snippets
```diff
@@ -56,23 +56,26 @@ ${1:class_name}.prototype.${2:method_name} = function`!p snip.rv = space_before_
...
 
-snippet fun "function (fun)" w
+snippet fun "function (named)" b
 function ${1:function_name}`!p snip.rv = space_before_function_paren(snip)`(${2:argument}) {
 	${VISUAL}$0
 }
 endsnippet
 
-snippet vf "Function assigned to var"
+snippet =fun "function (assigned to var)"
```
lpil: Could you put back the original trigger please? People will be using this.

idbrii: Sure, but what about anf vs fun for anonymous functions? Is that change okay? (If not, then I don't think this change has much left to offer.)

lpil: We should always aim to keep the existing triggers as people may be using them. We can add new ones :)

idbrii: I won't add `=fun` and I'll keep `anf` untouched (as a copy).

PR #1093, opened 2018-12-30
Fix rust trait/iter snippet name conflict

Avi-D-coder:
Fixes https://github.com/ncm2/ncm2-ultisnips/issues/15.
Expanding `trait` snippet resulted in `tra.iter()`.


lpil:
Thank you

Issue #1096, opened 2019-01-07
python snippets syntax error 


bkbncn:
syntax error at #1092  
 
`raise SnippetSyntaxError(filename, line_index, msg)
UltiSnips.snippet.source.file._base.SnippetSyntaxError: Invalid line '@unittest.skip(${1:skip_reason})' in .vim/bundle/vim-snippets/snippets/python.snippets:27 `

UltiSnips & python 3.7.2

PR #1098, opened 2019-01-08
Fix python.snippets syntax error

bkbncn:
Fix syntax error at PR #1092, which is mentioned at issue #1096 

Works with UltiSnips & python 3.7.2


lpil:
Sorry @bkbncn I didn't see this PR and have pushed a conflicting commit.

bkbncn:
@lpil No problem, never mind.

PR #1099, opened 2019-01-08
Add Drupal 8 UltiSnips.

jimafisk:
Based on https://github.com/dssiqueira/drupal-8-snippets-vs-code

### DIFF
--- a/UltiSnips/php-drupal8.snippets
+++ b/UltiSnips/php-drupal8.snippets
@@ -1,0 +1,2402 @@
+snippet hook_ajax_render_alter "Alter the Ajax command data that is sent to the client." b
+/**
+ * Implements hook_ajax_render_alter().
+ */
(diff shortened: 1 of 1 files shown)


lpil:
These snippets are numerous and the triggers are very long. It's quite different from what we have for other languages and frameworks.

Would it not be better to have a hook snippet and then to type in the name yourself?

jimafisk:
I figured the snippets would serve as a library for [Drupal hooks](https://api.drupal.org/api/drupal/core!core.api.php/group/hooks/8.2.x) so the user would know which hooks are available and the parameters that are used. Maybe that's not in the spirit of how UltiSnips are meant to work, but it was helpful for me. 

@lpil do you want to close for now and revisit if other people voice support for Drupal 8 hooks implemented this way?

lpil:
Let's do that, thank you. If there's a way we can help drupal users in a more conservative way that would be welcomed.

PR #1102, opened 2019-01-13
Add snippets for Phoenix web framework

josephan:
Phoenix is a popular web framework for the Elixir programming language.

I find myself often creating new controllers, views, and channels. These snippets will come in handy in web development.

### DIFF
--- a/snippets/phoenix.snippets
+++ b/snippets/phoenix.snippets
@@ -1,16 +1,16 @@
-snippet controller
-	defmodule ${0:app_name}Web.${1:controller_name}Controller do
-		use ${0:app_name}Web, :controller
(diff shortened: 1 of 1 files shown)


josephan:
To test you need to run `:UltiSnipsAddFiletypes phoenix` on an `*.ex` file.
How does `rails.snippets` get automatically loaded for rails projects but not for `*.rb` files?
I would like to do the same for this snippet.

lpil:
> How does rails.snippets get automatically loaded for rails projects but not for *.rb files?

I believe this is done by the vim-rails plugin if it finds a config/routes.rb (or something like that).

PR #1107, opened 2019-01-21
Add placeholders at end of CSS snippets

### LINKED ISSUE #1106: Should be able to tab cursor to be after snippet text
For example, CSS snippets like `c` for `color` should allow the user to tab the cursor to be *after* the closing semicolon; instead of this:

`color: #${0:000};`

this:

`color: #${1:000};${0}`

Otherwise, the cursor is stuck to the left of the semicolon in insert mode, so to start inserting after the semicolon you have to leave insert mode and move the cursor manually, by hitting `<Esc><Shift-A>`.

jsit:
Closes #1106

### DIFF
--- a/snippets/css.snippets
+++ b/snippets/css.snippets
@@ -1,50 +1,50 @@
 snippet . "selector { }"
 	${1} {
-		${0:${VISUAL}}
-	}
+		${2:${VISUAL}}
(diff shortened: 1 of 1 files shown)


lpil:
Thanks

PR #1109, opened 2019-01-27
Rust: Add dbg! macro snippet 'db'

phansch:
This was added with Rust 1.32 and is super useful compared to doing
just `print!` debugging.

See https://blog.rust-lang.org/2019/01/17/Rust-1.32.0.html#the-dbg-macro
for examples.


lpil:
What would you think of a trigger of `d`?

phansch:
Thank you for the fast response :+1:

PR #1115, opened 2019-03-10
remove space

heiderich:
between "\\emph{${1:${VISUAL:text}}}" and "${0}"

as I consider it annoying (especially for instance before a period or a comma). I wonder it those spaces should be removed in similar


lpil:
Thank you

PR #1116, opened 2019-03-11
Creates File and Function header snippets for Ultisnips

SethGower:
I spent some time creating these two snippets, most of which was used for making the second one. I based it off of what I learned reading through the python snippets in this repo, and figured I would share. The function one creates a function, and adds a header comment block, with each parameter having a space for a definition in the comment block, which expands as you add more. 
Demo


UltiSnips/c.snippets
```diff
@@ -96,4 +96,54 @@ snippet fund "function declaration" b
...
+if args:
+	for arg in args:
+		snip.rv += '\n' + '*' + ' '*19 + arg + ' - TODO'
+snip << 2
+`
+* Return:           $5
+* Error:            $6
+*****************************************************************************/
+${1}($2){
+${0}
```
lpil: I'm not very familiar with C, shouldn't this be indented?

UltiSnips/c.snippets
```diff
@@ -96,4 +96,54 @@ snippet fund "function declaration" b
...
+endglobal
+
+post_jump "if snip.tabstop == 0 : split_line(snip.tabstops[4].current_text)"
+snippet head "File Header" b
+/******************************************************************************
+* File:             `!p snip.rv = fn`
+*
+* Author:           ${2:`git config --get user.name`}  
+* Created:          `date +%m/%d/%y` 
+* Description:      ${4:I need a really long default here so I don't have to type so much !}
```
lpil: Could you remove the placeholder text here please? https://github.com/honza/vim-snippets#things-to-consider-when-contributing

Thank you

UltiSnips/c.snippets
SethGower: Yeah, I noticed that too. I fixed it. I will recommit

UltiSnips/c.snippets
SethGower: Which placeholder? the name or the Description, or both?

PR #1118, opened 2019-03-22
Disable all monkey patching done by RSpec.

majjoha:
This ensures that `describe` blocks are prepended with `RSpec` to fix specs where monkey patching is disabled. See https://relishapp.com/rspec/rspec-core/v/3-8/docs/configuration/zero-monkey-patching-mode for more information.

### DIFF
--- a/snippets/ruby.snippets
+++ b/snippets/ruby.snippets
@@ -653,6 +653,10 @@ snippet mthrow
 #     Rspec snippets     #
 ##########################
 snippet desc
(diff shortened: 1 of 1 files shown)


majjoha:
Sure. I decided to change the `desc` snippet rather than adding a new one as the substitution in the snippet indicates that it is supposed to be used for top-level `desc` blocks, and the `descm` snippet already fits other cases but I am happy to add an additional snippet if you still think it makes sense. :-)

lpil:
Yes please! As this is a commonly used snippet I'd like to avoid breaking changes :)

majjoha:
No problem! I've updated the PR with a new snippet instead.

PR #1126, opened 2019-05-10
Assume C# class, struct, interface match file name

idbrii:
Give a more useful default value for class, struct, interface snippets.
No one names their interface IInterface, but the filename might be
correct.


lpil:
Thank you

PR #1128, opened 2019-05-28
Add functional class snippet

chris542:
There was a missing snippet that a lot of people use in Visual Studio code :
Creating a functional component using rfc
I added rfc and also changed 
</div> to be whatever you type on ClassName so that it the component is not just a useless empty div

### DIFF
--- a/snippets/javascript/javascript-react.snippets
+++ b/snippets/javascript/javascript-react.snippets
@@ -42,11 +42,24 @@ snippet pt
 	static propTypes = {
 		${1}: React.PropTypes.${2:type},
 	}
+snippet rfc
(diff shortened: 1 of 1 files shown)


snippets/javascript/javascript-react.snippets
```diff
@@ -42,11 +42,24 @@ snippet pt
 	static propTypes = {
 		${1}: React.PropTypes.${2:type},
 	}
+snippet rfc
+	const ${1:ClassName} = (${2:props}) => {
```
lpil: Could we call this `ComponentName` rather than `ClassName`? Thanks

chris542:
Hey @lpil I made changes for this :)

Cheers!

PR #1146, opened 2019-09-12
tex: add tsub and tsup for textsubscript and textsuperscript

weber-s:
Hei! Thanks for all the snippets!

Here is a proposal for `tex` file `\textsubscript{}` and `\textsuperscript{}`.
I needed a lot super/sub-script, so I added it to my snippet, maybe it would be useful to include it.


weber-s:
Indeed, thanks for the comment!

PR #1149, opened 2019-09-19
added dart snippets for basic flutter function and UltiSnips

qazsxdcfv:
Hello guys I'm new to Git, Vim also Flutter, I found there is no dart ( in this case Flutter ) snippets for UltiSnips so I created one myself. It's just 5 snippets that i use commonly ( right now ) with Flutter as I learn more. Maybe this could help you guys, and I'm absolutely new to Git so please I'm sorry if there's anything wrong with pull request or such. :)

Thanks for considering my request

### DIFF
--- a/UltiSnips/dart.snippets
+++ b/UltiSnips/dart.snippets
@@ -1,0 +1,40 @@
+priority -50
+
+snippet im "Import Package" b
+import 'package:$1/$2.dart';
+endsnippet
+
+snippet imm "Import Material" b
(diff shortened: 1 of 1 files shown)


lpil:
Hi @qazsxdcfv ! Ultisnips can read the snipmate format, so we tend to only add snippets in the Ultisnips format if they make use of any special Ultisnips features. Would you mind porting these over to the snipmate format (presuming they do not already exist there). Thanks

PR #1150, opened 2019-09-21
Remove pre Rust 2018 edition snippets + Coc compatibility

Avi-D-coder:


Avi-D-coder:
They clash with any other e prefixed completion. By now I'm not sure there are any 2015 rust projects left.  Migration was supper easy.

lpil:
Which snippets clash with the same trigger? I can't see them.

Avi-D-coder:
Not snippets, Buffer and LSP completions.

lpil:
We only rename snippets in case of trigger collisions I'm afraid.

Avi-D-coder:
I'm suggesting they be removed, not only because they can clash, but because they encourage incorrect usage. See the [rust guide](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-extern-crate).

lpil:
I'm aware of the situation with Rust 2015 and 2018, I am a Rust programmer.

In this project we do not remove support for older styles so quickly as we want to support projects that have not upgraded. See the HTML and Javascript snippets for examples.

I'm going to close this now as we don't seem to be making progress. Thank you for your contribution.

PR #1153, opened 2019-09-26
snippets for vue-i18n

thet:
Two snippets to faster insertion of i18n translation statements:
More info:
http://kazupon.github.io/vue-i18n/guide/formatting.html
https://github.com/kazupon/vue-i18n


lpil:
Thanks

PR #1154, opened 2019-10-01
Adding some generic parentheses snippets.

jlconlin:
I've created a few automatic snippets for parentheses (and their cousins). It makes it easy to add a pair of parentheses and put something before, inside, and after it.

### DIFF
--- a/UltiSnips/all.snippets
+++ b/UltiSnips/all.snippets
@@ -107,4 +107,23 @@ snippet uuid "Random UUID" w
 `!p if not snip.c: import uuid; snip.rv = uuid.uuid4()`
 endsnippet
 
+
+#################
+#  Parentheses  #
+#################
+snippet () "Parentheses surround" A
+$1(${2:${VISUAL}})$0
+endsnippet
+
+snippet [] "Bracket surround" A
+$1[${2:${VISUAL}}]$0
+endsnippet
+
+snippet {} "Brace surround" A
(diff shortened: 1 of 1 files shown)


UltiSnips/all.snippets
```diff
@@ -107,4 +107,23 @@ snippet uuid "Random UUID" w
...
+$1[${2:${VISUAL}}]$0
+endsnippet
+
+snippet {} "Brace surround" A
+$1{${2:${VISUAL}}}$0
+endsnippet
+
+snippet <> "Angle surround" A
+$1<${2:${VISUAL}}>$0
+endsnippet
```
lpil: Do these need to be in the ultisnips format? It looks like they could be written in the snipmate format

jlconlin: I suppose they could. I'm not as familiar with the snipmate format. Is it preferred for them to be in the snipmate format?

lpil: Yes, the snipmate format is more widely supported. :)

jlconlin: Does snipmate support automatic trigger like ultisnips? That's an important feature here.

If this was to be done for snipmate, what file would it go under? There isn't an `snippets/all.snippets` file.

lpil: Ah yes! Thanks :)

PR #1156, opened 2019-10-04
Remove redundant whitespaces from Python todo

### LINKED ISSUE #1145: TODO snippet inserts too many spaces in Python
I just noticed using the `todo` snippet inserts the following text in Python
```
#_ TODO:  <10-09-19, Name> #_
```
I spot two redundant spaces here, marked as `_` for clarity.
I guess the issues originates from `get_comment_format()[0]`, which probably returns `#_`.

sbmueller:
This fixes #1145 .
I chose to remove leading whitespaces in `all.snippets` todo-snippet, since whitespaces should IMO be managed by the language's comment format. There might be languages that discourage leading spaces in comments. In Python, this lead to two leading whitespaces.
In pythonx's `get_comment_format()`, I removed trailing whitespaces from the `end_line` comment format, avoiding trailing whitespaces to be inserted into the code.

### DIFF
--- a/pythonx/vimsnippets.py
+++ b/pythonx/vimsnippets.py
@@ -69,7 +69,7 @@ def get_comment_format():
     commentstring = vim.eval("&commentstring")
     if commentstring.endswith("%s"):
         c = commentstring[:-2]
-        return (c, c, c, "")
+        return (c.rstrip(), c.rstrip(), c.rstrip(), "")
     comments = _parse_comments(vim.eval("&comments"))
     for c in comments:
         if c[0] == "SINGLE_CHAR":


lpil:
This fixes Python but breaks other formats such as Javascript as now it is missing a space. Can we support both?

sbmueller:
Following my suggestion, `get_comment_format()` should return the correct comment format for each language. If Javascript is using `'// '` as comment format, this should be returned from `get_comment_format()`. Is ist just returning `'//'` currently? Then, IMO this is inconsistent: For Python returning whitespace, for Javascript not, when in both cases it would be correct.

I see two options:
1. Hotfix: Strip all whitespaces from comment formats and handle whitespaces in the snippets
2. Leading whitespaces are language dependent and therefore should be provided by `get_comment_format()`. Since they originate from vim's python package, it could be an issue there, which needs to be fixed.

lpil:
Isn't `get_comment_format` defined in `pythonx/vimsnippets.py`? I'm not sure we'd need to make a change to another project.

sbmueller:
Yes, however the method basically uses vim to get the comment format:
```python
commentstring = vim.eval("&commentstring")
print(f"'{commentstring}'", flush=True)
```
yields: `# %s` for Python and `//%s` for Javascript. I've changed the PR to strip whitespaces from the comment format, leaving the whitespacing up to the snippets themselves.

PR #1162, opened 2019-10-11
Add file name for vim in UltiSnips

kanneg881:
it is convenient to type file name.


lpil:
Could we instead have this in the snipmate format?

kanneg881:
@lpil no problem I will modify it and add new pull request.

PR #1168, opened 2019-10-31
Refactor complete function to allow 'fuzzy' match

ashfinal:
This PR will allow 'fuzzy' match and coexist well with pum.

complete

### DIFF
--- a/pythonx/vimsnippets.py
+++ b/pythonx/vimsnippets.py
@@ -1,25 +1,25 @@
 """Helper methods used in UltiSnips snippets."""
 
-import string, vim
+import string, vim, re
 
 def complete(tab, opts):
     """
-    get options that start with tab
+    get options that match with tab
 
     :param tab: query string
     :param opts: list that needs to be completed
 
-    :return: a string that start with tab
+    :return: a string that match with tab
     """
-    msg = "({0})"
-    if tab:
-        opts = [m[len(tab):] for m in opts if m.startswith(tab)]
-    if len(opts) == 1:
-        return opts[0]
-
-    if not len(opts):
-        msg = "{0}"
-    return msg.format("|".join(opts))
+    el = [x for x in tab]
+    pat = "".join(list(map(lambda x: x + "\w*" if re.match("\w", x) else x,
+                           el)))
+    mats = [x for x in opts if re.search(pat, x, re.IGNORECASE)]
+    if len(mats) == 0 or str.lower(tab) in list(map(str.lower, mats)):
+        return ""
+    cads = "|".join(mats[:5])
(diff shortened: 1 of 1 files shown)


lpil:
Nice!

idbrii:
@ashfinal 
How does this fuzzy match work? If I typed `sb`, I get `StackedBar` as a result, but no way to select it (Ctrl-n) requires matching prefix.

ashfinal:
@idbrii I can't believe I missed your message! 😨 
This PR actually didn't use any fuzzy algorithm, just something like regex match. :) As for your question, `Ctrl-n` requiring matching prefix, well, it depends on your completion engine(and fuzzy algorithm). Basically, you need to make sure the option appears in your popup menu before select it. For instance, since I use `coc.nvim`, when need `StackedBar`, I type `Sb` instead of `sb`, because capital letters have more weight according to coc's fuzzy algorithm.

It is a little tricky and sometimes may confuse people, I have to admit. Because it demands you knowing more about your completion engine. But the good thing is it could coexist with most completion engines, and vim's built-in completion mechanism.

If anyone has better idea or implementation, I'd like to discuss further. :)

idbrii:
I guess that means you have an imap for `<C-n>`? Would it make sense to detect coc and other fuzzy searchers (`'did_coc_loaded' in vim.vars and vim.vars['did_coc_loaded']`) and only enable fuzz for those?

Otherwise, it's a bit confusing to see options that you can't select if you're not using coc.

ashfinal:
The imap for <C-n> is built in Vim, unless some plugin overrides it, see `:help i_Ctrl-n`. 

If you don't use any completion engine, you can type `St`, then <Ctrl-n>, select the choice that you want. In other completion engines, the typed letters may vary. 

Maybe the most intuitive way to do this(without conflicting with any completion engine) is put the choices into popup menu in the first place, which basically means write another minimal completion engine with fuzzy algorithm, IMO, that demands more work...

Still, I'm quite insterested in completion engine. I will see what I can do If I get time.

idbrii:
> If you don't use any completion engine, you can type St, then <Ctrl-n>, select the choice that you want. In other completion engines, the typed letters may vary.

My point is that if you type `bar`, the choices narrow down to the one that you want, but you can only Ctrl-n to select it if you have coc. For those without coc (like me), this looks like a bug.

ashfinal:
Yeah, it's not the most intuitive way to do things. I hope to get time and energy to work on it. Or someone else could.

mperry2:
Why are you intentionally hiding completion options? That makes this completion function nearly useless if you have more than 5 items to choose from.

PR #1176, opened 2019-11-24
Fix some rails snippets that were failing because of a closing brace

apux:
Some snippets had a closing brace escaped, and that caused them to show incorrectly.
It also fixes a hash syntax.


apux:
Before the change:
vokoscreen-2019-11-23_21-27-36

After:
vokoscreen-2019-11-23_21-26-16

Issue #1178, opened 2019-11-26
Do we need more snippets for Fortran?


yhu266:
I have a small collection of Fortran snippets [here](https://github.com/yhu266/vim-fortran-snippets). Some of them do include UltiSnips features.

My question is, should I contribute them to this repo? Or just leave it as separated? How can we tell whether a snippet is generally accepted or of personal flavor?

Thanks!

lpil:
New snippets are accepted! We try not to clash with any existing triggers though

yhu266:
> New snippets are accepted! We try not to clash with any existing triggers though

That's great! I'll make sure the new ones are backward compatible!

PR #1179, opened 2019-11-27
Fortran snippets with UltiSnips feature

yhu266:
#1178

### DIFF
--- a/UltiSnips/fortran.snippets
+++ b/UltiSnips/fortran.snippets
@@ -1,0 +1,267 @@
+priority -50
+
+# Common
+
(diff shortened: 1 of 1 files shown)


yhu266:
Working on it! One more question:

I usually set up `UltiSnips` with
```viml
let g:UltiSnipsEnableSnipMate = 0
```
which ignores all `snipMate` snippets. So should I keep the duplicates under `UltiSnips` directory in case people have similar settings?

lpil:
Our policy is not to have duplicates between the two formats I'm afraid.

PR #1186, opened 2020-01-12
Remove undefined "pipefail" from shell snippet

Sh3Rm4n:
For POSIX shells pipefail is undefined.
bash on the other hand has defined pipefail

Reference:

https://github.com/koalaman/shellcheck/wiki/SC2039

### DIFF
--- a/snippets/sh.snippets
+++ b/snippets/sh.snippets
@@ -4,10 +4,10 @@ snippet #!
 
 snippet s#!
 	#!/usr/bin/env sh
-	set -euo pipefail
+	set -eu
 
 snippet safe
-	set -euo pipefail
+	set -eu
 
 snippet bash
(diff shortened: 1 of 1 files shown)


lpil:
Doesn't the -o also need to be removed? Could we also remove it from the other snippets that use it? Thanks :)

Sh3Rm4n:
Oh yeah, I forgot that o is actually the option flag with `pipefail` as parameter.

Thanks for that :)

I've only found this case, where `sh` is explicitly set and `pipefail` is used.
Will have another look.

Sh3Rm4n:
Maybe this one:

https://github.com/honza/vim-snippets/blob/977cc05194bdac51970bf83d0fe6fafc297012de/snippets/sh.snippets#L9-L10

I'm not sure though

PR #1195, opened 2020-02-20
Added julia test snippets

apieum:
\@t for \@testset with \@test and \@tt for \@testset with \@test_throws


apieum:
warn, info methods are not working, I suppose it was for an old version of julia, now it's macros @info, @debug, @error, and @warn, I'll propose another pr after this one to fix it, if you want the 2 in one let me know. thank you

PR #1207, opened 2020-04-11
Update install instructions with vimplug

waltsims:
Added simple example for vimplug installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ ActivateAddons vim-snippets snipmate
 Installation with [vimplug](https://github.com/mahyarap/vimplug): 
 
 ```vim
-" install to ~/.vim/snippets
+" install vim-snippets
 Plug 'honza/vim-snippets'
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,13 @@ supports simple dependency management. E.g. all you need is this line in your
 ActivateAddons vim-snippets snipmate
 ```
 
+Installation with [vimplug](https://github.com/mahyarap/vimplug): 
+
+```vim
+" install to ~/.vim/snippets
+Plug 'honza/vim-snippets', { 'dir': '~/.vim/snippets' }
```
lpil, line 104: Why this specific directory?

waltsims: The default install directory with vimplug is `~/.vim/bundle/vim-snippets/`.

I found that by default the vim-snippets were not found here when using Ultisnips so I installed the plugin under `~/.vim/snippets` to ensure they were found in the run-time path.

waltsims: Another possibility I just found is to update the Ultisnips search path which allows one to use the default install path. If you think it would be better, I can change the entry accordingly.

lpil: I use vim plug and the default directory works with ultisnips for me, I'm curious as to why it does not for you.

waltsims: Me too. I will update the README to default and see if I can find my issue.

PR #1208, opened 2020-04-14
added ejs.snippets file

douglas86:
added a simple ejs.snippets file,
won't you please tell me what you think.


douglas86:
Added another file react hooks called javascript_react.snippets

PR #1220, opened 2020-05-02
new updates for javascript_react snippets file

douglas86:
Fixed rfc snippet and included an img and a div tag to the mix


lpil:
Hi there. Thanks for your contribution but I don't think we should be including imports that are not used in snippets, and if html snippets are desired best update your editor to load the full set of HTML snippets for JavaScript rather than duplicate some of them in this snippet file.

Thank you

PR #1226, opened 2020-05-15
Add missing angle bracket for markdown Disclosure snippet

unicell:


unicell:
cc @BaksiLi who originally added the snippet

I might be wrong, but don't you think there's a missing angle bracket here?

Issue #1213, opened 2020-04-22
provide some snippets like this?


Freed-Wu:
```
/*
 * %
 */
```

some code has this kind of comment. can it provide this snippets?
thanks!

lpil:
A PR will be accepted for this :)

PR #1228, opened 2020-05-29
Close #1212

### LINKED ISSUE #1212: provide some snippets?
i notice `all.snippets` has provided some snippets like box.
```
""""""
"  %  "
""""""
```

% is the position of cursor, " is the comment (e.g. vim)

can it provide some snippets like

```
" ┌──┐
" │  %│
" └──┘
```

Freed-Wu:
image

### DIFF
--- a/UltiSnips/all.snippets
+++ b/UltiSnips/all.snippets
@@ -52,6 +52,13 @@ snippet boxx "A nice box with a nice border" b
 $0
 endsnippet
 
+snippet comment "comment style like javadoc" b
+`!p snip.rv = get_comment_format()[0]`
+ `!p snip.rv = get_comment_format()[1]` ${1:${VISUAL:content}}
+ `!p snip.rv = get_comment_format()[2]`
+$0
+endsnippet
+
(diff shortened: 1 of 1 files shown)


lpil:
Thanks for this!

I tried these snippets on Python, C, Ruby, JavaScript, Java, Haskell and Lisp and it didn't look right on any of them. I think it only works if the comment format is `/* text */`. Could we make it work on all comment styles? Thanks

Freed-Wu:
about #1212, this is a test.vim:
```{.vim}
for s:lang in ['python', 'vim', 'c', 'java', 'ruby', 'perl']
	execute 'setlocal filetype=' . s:lang
	execute 'normal }oboxx	'
endfor
```

this is a test result:
image

about #1213, i test it
image

but in fact, i wanna get this in julia
image

but it looks a bit hard...

do you think there are other problem?

lpil:
Most of those should not have the comment at the end of the line for `boxx`, and they should not have the extra  space before the comment on the 2nd and 3rd line for `comment`

Freed-Wu:
it is a little hard...

Issue #1227, opened 2020-05-19
box snippet for rust buffers is overshadowed


brainplot:
The `box` snippet, as defined here https://github.com/honza/vim-snippets/blob/087d3e7c72912baeb6b1d7ba626e61d50092c848/snippets/rust.snippets#L241-L242
which should expand to a `Box<T>` declaration is overshadowed by the box comment.
If I try to expand it with `box<tab>` I'll get this instead:
```
///////////////
//  content  //
///////////////
```

PR #1229, opened 2020-06-02
Improve/add some Rust snippets

brainplot:
- I added a final `p` to the `box` snippet which was overshadowed by the global `box` snippet, as I reported in issue #1227.
- `if let (...)`,  `while let (...)` and `use` are fairly common constructs in Rust, so I added snippets for those.
- Snippets that expand to `struct` and `trait` definitions I found are a bit inconvenient to use when there are generic types involved, because you'd need to go back and add them where needed, which in a way takes away those time savings you want when using a snippet, imo. This is especially a problem in Rust where you need to specify the generic type in more than one spot. For that reason, I added an additional tab stop that the user can quickly delete and move on if a generic type is not needed at all and, if inserted, it will be duplicated everywhere.

I've been testing those snippets for a few days and they seem to work for me. I copy-pasted them to make this PR but haven't tested them. They should work but let me if there's a copy-paste error somewhere.

### DIFF
--- a/snippets/rust.snippets
+++ b/snippets/rust.snippets
@@ -122,6 +122,10 @@ snippet ife "if / else"
 	} else {
 		${0}
 	}
+snippet ifl "if let (...)"
+	if let ${1:Some(${2})} = $3 {
+		${0:${VISUAL}}
+	}
 snippet el "else"
 	else {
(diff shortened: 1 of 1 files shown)


brainplot:
Thanks for merging this!

PR #1232, opened 2020-06-27
add snippet for gitcommit co-author directive

AlexRiina:
This directive will make the commit display as having multiple authors in github.

https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

### DIFF
--- a/snippets/gitcommit.snippets
+++ b/snippets/gitcommit.snippets
@@ -1,0 +1,2 @@
+snippet co
+	Co-authored-by: ${1} <${2}>


lpil:
Oops, it looks like the snippet syntax is invalid. Could you fix that please? See the other files in the snippets directory

AlexRiina:
Woops, thanks for pointing that out. I've moved the snippet to the UltiSnips directory since I was using that syntax.

lpil:
Please use the snipmate format, it is the universal format that all our users can use.

AlexRiina:
Ah ok, moved back to snippets/ and re-implemented in snipmate

PR #1234, opened 2020-07-15
fix clojure snippets

gs:
I've noticed that after expanding clojure snippets like `defn`, `defm`, `defp`, `defr`, `defty`, `doseq`, `fn`.... I cannot move the coursor to next placeholder.  It just stucks on the first placeholder

I was able to fix this problem after adding the name attribute next to index - `code`.

This is only replates to snippets that have placeholder with names and the last placeholder does not have a name attribute.

This is tested with my configuration which is `neovim` + [coc-snippets](https://github.com/neoclide/coc-snippets) and vim-snippets.


honza:
Thanks!

PR #1249, opened 2020-08-27
Added a bunch of math tex snippets for UltiSnips

paniash:


lpil (commented):
It looks like these don't use any special Ultisnips features so could we move them to the snipmate format in the snippets directory? The Ultisnips plugin can also use that format.

paniash:
Pardon me since I'm not aware of this but what special features of UltiSnips is incompatible with snippets?

paniash:
Ok, I'll port it to the snipmate format and let you know soon.

lpil:
Thank you :)

paniash:
Is it alright if my snippets are still present in the UltiSnips folder or do I have to remove them?

lpil:
Remove them please, ultisnips can load the snipmate format without any extra code :)

paniash:
@lpil I've just pushed the changes as you suggested. Can you please take a look and let me know if it's ok?

lpil (approved):
Thank you

paniash:
Thanks! @lpil

PR #1251, opened 2020-09-04
Add const async arrow function JavaScript snippets

nomasprime:


lpil (approved):
Thank you

nomasprime:
No problem @lpil 

Related question: looks like aliases still haven't found their way into UltiSnips so is the best workaround to fork the definitions, ie, this repo?

PR #1256, opened 2020-09-24
Speed up the startup

leiserfg:
Now git is called only 2 times and using `command` makes the shell avoid resolving aliases. That decreases the startup time of the plugin by around 300ms on my laptop.
Anyways IMHO those vars should be on the README, because, even with my fix, this plugin makes vim to take noticeable time to start if they are not set.

### DIFF
--- a/plugin/vimsnippets.vim
+++ b/plugin/vimsnippets.vim
@@ -4,23 +4,24 @@ endif
 let b:done_vimsnippets = 1
 
 " Some variables need default value
-if !exists("g:snips_author") && executable('git')
-    let g:snips_author = trim(system('git config --global user.name'))
+let s:has_git = executable('git')
+
+if !exists("g:snips_author") && s:has_git
+    let g:snips_author = trim(system('command git config --global user.name'))
 else
     let g:snips_author = "yourname"
 endif
 
-if !exists("g:snips_email") && executable('git')
-    let g:snips_email = trim(system('git config --global user.email'))
+if !exists("g:snips_email") && s:has_git
+    let g:snips_email = trim(system('command git config --global user.email'))
 else
     let g:snips_email = "yourname@email.com"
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you. Any further improvements to the code or docs are very welcome

ashfinal:
This commit causes my Windows 10 machine to popup a window asking whether to install some "NTVDM" feature or not every time I launch Neovim-qt, which I dare to say, is quite overwhelming and suspicious in a normal user's view.

Fondue_CIRtiRx7Mo

At first, I thought it was Neovim-qt, maybe they changed something unconsciously. Well, they did not. It turns out it's this commit: the system() function trying to invoke a 16-bit application `command`, which was removed long time ago.

```
Program 'COMMAND.COM' failed to run: %1 is a 16-bit application. You do not have permissions to execute 16-bit applicat
ions. Check your permissions with your system administratorAt line:1 char:1
+ command
+ ~~~~~~~.
At line:1 char:1
+ command
+ ~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed
```

Could you please fix this or at least revert the commit? Also, it seems feasible to set `g:snips_author` and `g:snips_email` in your vimrc, so the system() function doesn't have to run, thus speeding up the startup. Thanks a lot, and cheers!

lpil:
I think you're right, we've not got this right yet. I'll revert these changes.

leiserfg:
Then just remove the `command` part, I only checked `bash`, `dash`, and `fish` shells on linux not `cmd` on windows.

lpil:
Good idea, thanks. Let's try that

ashfinal:
@lpil Cool, it's ok now.

PR #1259, opened 2020-09-29
"method" snippets added to typescript as public, private and protected.

mthnglac:
"public" is a default modifier in typescript; but I added it because I believe it helps the code to be more readable.


lpil (approved):
Thank you

PR #1263, opened 2020-10-01
if-else-if condition snippet added for javascript

mthnglac:
snippet for if-else-if condition.

### DIFF
--- a/snippets/javascript/javascript.snippets
+++ b/snippets/javascript/javascript.snippets
@@ -61,6 +61,15 @@ snippet ife "if (condition) { ... } else { ... }"
 	} else {
 		${2}
 	}
+#if-else-if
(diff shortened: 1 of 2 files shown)


snippets/javascript/javascript.snippets
```diff
@@ -61,6 +61,15 @@ snippet ife "if (condition) { ... } else { ... }"
...
 	}
+#if-else-if
+snippet ifelif
+	if (${1})
+		${2:${VISUAL}}
+	else if (${3})
+		${0:${VISUAL}}
+snippet eif
+	else if (${3})
+		${0:${VISUAL}}
```
lpil, line 72: Could we have {}s to match the other if else snippets?

PR #1265, opened 2020-10-02
haskell: pre- and post- dot tabstops on import directives

Aster89:
`Data.Text` is only one of the several modules which we want to import/import from/import with a different name, therefore this PR 
  - modifies the `imp` UltiSnips snippet to that two other placeolders are added for the text before and after the dot
  - adds the `imp` with similar logic for the simple import
  - add `imp2` as well, but since the last argument/tabstop is between parenthesis, the final tabstop is placed after the parenthesis.

Let me know if you like it.

### DIFF
--- a/UltiSnips/haskell.snippets
+++ b/UltiSnips/haskell.snippets
@@ -1,5 +1,13 @@
 priority -50
 
+snippet imp "Simple import"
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

Issue #1266, opened 2020-10-02
Duplicate sum in latex snippets


vcoyette:
Hi,
There are two snippets 'sum' in snippets/tex.snippets. They are not exactly the same, but very similar, I think one of them should be renamed (summ for the summation?) or deleted.

lpil:
I think you are right. Please open a pull request fixing the issue if you have the time. Thank you

vcoyette:
Sure, which one do you prefer ? Delete or rename the second one ?

lpil:
Rename sounds good to me. I don't mind what the name is though!

PR #1268, opened 2020-10-03
fix duplicated sum snippet in tex

vcoyette:
Rename second sum snippet in tex.snippets as discussed in issue #1266.


lpil (approved):
Thank you

PR #1276, opened 2020-10-10
Use initial-scale in html snippet.

UlisseMini:
[Motivation](https://css-tricks.com/probably-use-initial-scale1/) plus the viewport snippet already has this


lpil (approved):
Nice, thank you

PR #1277, opened 2020-10-12
created public/private/protected field snippet for typescript

mthnglac:
I used the "fi" suffix for the word field. It does not conflict with existing snippets either.


lpil (approved):
Thank you

PR #1278, opened 2020-10-13
Update Ocaml documentation and comment snippets

blurrcat:
According to the official Ocaml documentation guideline, documentations should be wrapped in `(**` and `*)` whereas comments should be wrapped in `(*` and `*)`: https://ocamlverse.github.io/content/documentation_guidelines.html


lpil (approved):
Thank you

PR #1279, opened 2020-10-13
Fix error when using box snippet in HTML

whonore:
Previously expanding the `box` snippet in an HTML file raised the following error:
```python
Following is the full stack trace:
Traceback (most recent call last):
    ... 
    sline = b + m + bwidth_inner * m[0] + 2 * m[0]
IndexError: string index out of range
```

This is because `comments` is `s:<!--,m:    ,e:-->`, and `make_box` strips whitespace so `m` is an empty string. The proposed fix now makes a box like:
```html
<!--
  content
-->
```
which is pretty ugly, but I'm not sure how to do better without making a special case for HTML.


lpil (approved):
Thank you!

PR #1281, opened 2020-10-18
Vim: remove useless comment from `au` snippet

Aster89:
I don't know where this comment comes from, nor what it is for, but I bet it's completely useless.


lpil (approved):
Thank you

Issue #1282, opened 2020-10-19
Wrong Java "find" methods snippets


alexghergh:
Hello,

I feel like these snippets are wrong:

```
snippet findall
	List<${1:listName}> ${2:items} = ${1}.findAll();
snippet findbyid
	${1:var} ${2:item} = ${1}.findById(${3});
```
I think they should be:

```
snippet findall
	List<${1:listName}> ${2:items} = $1.findAll();
snippet findbyid
	${1:var} ${2:item} = $1.findById(${3});
```

Notice that the second one doesn't have braces around the second `$1`.

The first one puts you straight to the second `$1`. I do not know why that works like that, but that's what happens.

If someone can confirm this is indeed a typo/issue/bug, I would be happy to take a look at the whole Java file and see if I can find other occurences of this.

lpil:
I think you're right. Would you mind making a PR to fix this? Thank you

PR #1283, opened 2020-10-20
Fix snipmate/java-snippets "find" snippets

alexghergh:
Fix an issue where some snippets would not work as intended due to erroneous braces, i.e.: ${1} instead of $1.

This pull request fixes issue #1282.


lpil (approved):
Thank you!

PR #1285, opened 2020-10-22
Update twig snippets

eirabben:
Updated twig snippets with a complete set of tags.


lpil (commented):
Looks great! It seems the `bl` and `ext` snippets have been deleted. Could we put those back? People may be using them already

eirabben:
`bl` and `ext` have been added.

lpil (approved):
Fab! Thank you

PR #1289, opened 2020-10-29
Remove the charset="utf-8" attribute of <script>

sedrubal:
According to <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-charset> using `charset` is deprecated and useless.


lpil (approved):
Nice! Thank you

PR #1296, opened 2020-11-25
C++: improved modern for loop

Aster89:
As a follow up to #1275, I've improved the snippet in such a way that it automatically generates the _condition_ for _init-statement_ matching these forms (see the demo below)

  - `begin(whatever)`,
  - `std::begin(whatever)`,
  - `boost::begin(whatever)`,
  - `whatever.begin()`,
  - `whatever->begin()`.

If this PR is merged, in the future I plan to make another (possibly last) improvement to handle a backward loop (as in, if you enter an `end`-like thing in the _init-statement_, then the condition adapts to the corresponding `begin`).

[asciicast](https://asciinema.org/a/EQ6kENjat6yp3vlsypFPJNAGt)

### DIFF
--- a/UltiSnips/cpp.snippets
+++ b/UltiSnips/cpp.snippets
@@ -31,7 +31,7 @@ endglobal
 #                            TextMate Snippets                            #
 ###########################################################################
 snippet forc "general for loop (for)"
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks!

PR #1297, opened 2020-11-26
C: Modern improved for - loop and modern printf

TornaxO7:
Added two of my favourite snippets :)

`printf` demonstration:
GIF

`for` - loop demonstration:
GIF

### printf
The printf snippet counts the amount of `%` and add the `, ` by the amount of the percentages.

### for - loop
It will always take the third last word of the first argument of the loop. For example
```c
for (int a=0)
```
`a` will be picked up and will be added to the third argument:
```
for (int a=0; ; a++)
```

### DIFF
--- a/UltiSnips/c.snippets
+++ b/UltiSnips/c.snippets
@@ -1,7 +1,33 @@
 ###########################################################################
 #                            TextMate Snippets                            #
 ###########################################################################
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1304, opened 2020-12-26
Add maketitle command snippet using mkt

Mikadifo:
I was missing basic and simple commands snippets like maketitle


lpil (approved):
Thanks

PR #1305, opened 2020-12-26
Add documentclass snippet with and without options

Mikadifo:
Whit options: dclo -> \documentclass[options]{class}
Whitout options: dcl -> \documentclass{class}


lpil (approved):
Nice, thanks

PR #1309, opened 2021-01-07
s/assert_raise/assert_raises/

paddor:
Minitest is now the standard and Test::Unit also provides #assert_raises


lpil (approved):
Thank you

PR #1310, opened 2021-01-11
Create snippets for puppet: type alias, lambda, lookup and trocla

lelutin:
These are convenience methods for some of the puppet syntax that's been added since the snippets were created.

The lambda snippet defaults to `each` since it's the most commonly called lambda, but it can be used for any other like map, filter and reduce.

The trocla function is not built into puppet but it is a convenient tool for automating management of certain passwords and other types of sensitive material. It differs from the more commonly used hiera-eyaml in that sensitive data is generated during puppet runs if missing and can be refreshed after some time has elapsed, providing automatic password rotation.

### DIFF
--- a/UltiSnips/puppet.snippets
+++ b/UltiSnips/puppet.snippets
@@ -8,11 +8,13 @@ global !p
 import vim
 import os.path
 def get_module_namespace_and_basename():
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1312, opened 2021-01-17
C++: not a plain C main

Aster89:
Isn't this a better version as regards modern C++?

### DIFF
--- a/UltiSnips/cpp.snippets
+++ b/UltiSnips/cpp.snippets
@@ -30,6 +30,13 @@ endglobal
 ###########################################################################
 #                            TextMate Snippets                            #
 ###########################################################################
+snippet main
+int main()
+{
+	${0}
+}
+endsnippet
+
 snippet forc "general for loop (for)"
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks

dragonxlwang:
The added snippet does not have `int argc, char *argv[]` cmd args? Before, the fallback c snippet has those args added https://github.com/honza/vim-snippets/blob/master/snippets/c.snippets#L4-L8

dragonxlwang:
pull #1327

Aster89:
There's a difference between C and C++. Most of the times you don't need `argc` and `argv` in a C++ codebase, and definitely not the `return` statement.

dragonxlwang:
Why's that? could you share some reference of that? Without argc/argv there's no way that you can pass cmd line args to the executable which imho has nothing to do with c/c++... 

as an example https://github.com/gflags/gflags is used in commonly and it relies on those cmdline args

Aster89:
I'm not say it's never needed. I'm just saying that it's not always needed. So why not providing two snippets instead of one?

dragonxlwang:
It would be okay to have an optional main snippet, e.g. main_no_args, for your proposal. 

Again, I think that most of cpp main still requires the args and it's best to have the default behavior with the args/argc included

Aster89:
Find, then let's pick another word to trigger the non-`args` snippet. But `main_no_args` seems long. What about `mainc`, or `mainp`, or `mainn`, or something this short?

dragonxlwang:
mainn would be fine, the same as the c version: https://fburl.com/h2o49s1b

Aster89:
In this case you can probably take the snippet I wrote and rename it to `mainn`, so the `main` is taken from C.

dragonxlwang:
Before the #1312, both main and mainn was from C, so if there's no special treatment for cpp, then it would be easier to just unland #1312?

Aster89:
Yeah, but none of them was
```
int main()
{
    ${0}
}
```

PR #1318, opened 2021-01-30
`END` to close an `augroup` should be capitalized

Aster89:
`:h :aug` for reference.

I discovered this because I was not able to jump between beginning and end of an auto group.


lpil (approved):
Thank you

PR #1327, opened 2021-02-23
make cpp main accepting args/argv

dragonxlwang:


dragonxlwang:
this is to make the changes #1312 behavior as cpp main

lpil (approved):

PR #1328, opened 2021-02-24
Added cuda snippets

luisbaldissera:
Added some snippets for CUDA language:

- Kernel definition and Kernel call
- Device function definition
- Shared memory declaration
- Thread synchronization
- `threadIdx`, `blockIdx` and `blockDim` (and index calculus based on it)


lpil (approved):
Thank you

PR #1329, opened 2021-03-03
Suggestions for fortran.snippets

tammasloughran:
The default Fortran snippets left me wanting more. Here are some suggestions.

Regarding the original procedure snippet: that wasn't how the procedure statement works in Fortran. It's just used as declarations in statements.

There are some naming changes to make it feel more like tab completion/distinguish from intrinsic functions.

Cheers,
Tam

### DIFF
--- a/snippets/fortran.snippets
+++ b/snippets/fortran.snippets
@@ -12,6 +12,11 @@ snippet modu
 	module ${1:modulename}
 		${0}
 	end module $1
+# submodule
+snippet submod
+	submodule (${1:parent}) ${2:submodulename}
+		${0}
+	end submodule ${2}
 # Procedure declaration
 snippet proc
 	procedure (${1:procname}) :: ${0}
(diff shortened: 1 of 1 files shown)


tammasloughran:
The changes to coding style like the spaces between the statements (eg. The open statement) and the parentheses options are a matter of tase. You can take them or leave them I guess; I forgot I put them in there.

lpil:
Would it be possible to keep some of the triggers the same as before? Some people may be using them and I wouldn't want to mess up their workflow

tammasloughran:
Sure. You can always duplicate the snippets to have multiple triggers.

tammasloughran:
The only thing I feel strongly about is print and write. pr should be print *, and write should be the write statement. Having pr -> write is nonsensical because they are not the same thing. print is only intended to be used for displaying strings on the standard out. write is used for writing data to files and variables (coincidentally it can also be used to write to standard out).

PR #1331, opened 2021-03-09
lua: Add assert

idbrii:
Standard lua assert message. Same in [lua5.1](https://www.lua.org/manual/5.1/manual.html#pdf-assert)  as [lua5.4](https://www.lua.org/manual/5.4/manual.html#pdf-assert) 

msg is optional, so hide the ", " if it's omitted.


lpil (approved):
Thank you

PR #1332, opened 2021-03-10
Add the 'autocmd!' command to clean the autocmd in the group

mdedonno1337:
This will prevent to add the already loaded autocmd when reloading the
vim configuration. The 'autocmd!' command remove the content of the
augroup at load.

https://learnvimscriptthehardway.stevelosh.com/chapters/14.html


lpil (approved):
Thank you

PR #1333, opened 2021-03-10
Add the augroup with visual placeholder snippet

mdedonno1337:
This will place the visual selection in the correct place in the `augroup` for vim files.


lpil (approved):
Thank you

PR #1334, opened 2021-03-12
Update html snippet lang from es to en

sevensidedmarble:
Most of the other lang examples use en (english) but this one uses es (spanish). Probably a mistake?


lpil (approved):
Thank you

Issue #1337, opened 2021-03-24
remove default value "body" for conventional commits


FabianUntermoser:
Hi there,
first thank you for this addition. I agree a good commit style should be practised.
However I noticed a small user experience problem, when using the new conventional commit templates.

In small projects, most of the time, I create one-line commit messages.
In the commit window (I use fugitive), I typically write `feat<TAB>` to expand the snippets.
However, as there is a default value "body" on every snippet, I additionally have to tab through and delete the body text if I only want a single line commit.

The problem would be solved, if you would keep the jump to the body section but leave the default value blank.
Would appreciate it, thanks!

lpil:
That sounds like a good improvement. Would you be up for making a pull request?

FabianUntermoser:
Alright I'll look into it.

FabianUntermoser:
@lpil Hi! I created a PR. Mind looking into it?

lpil:
Fab, thank you!

FabianUntermoser:
No problem :)
First PR here we goo :tada:

PR #1338, opened 2021-03-24
refactor(gitcommits): remove default value in conventional commit snippets

FabianUntermoser:
Fixes https://github.com/honza/vim-snippets/issues/1337


lpil (approved):
Great stuff. I especially like the use of VISUAL

PR #1339, opened 2021-03-25
Fix useS snippet in UltiSnips/javascript_react

Farzat07:
Previously, the .title() function was used to capitalize the first item.
However, this function doesn't maintain the capitalization of letters in
the middle. Therefore, a custom function was created to effectively
convert camelCase to PascalCase.


lpil (approved):
Thanks

PR #1342, opened 2021-04-01
Include arg type in google-style docstrings

nathanalderson:
Results in something like this:

```python
def function(i: int, s: str, o: MyObject):
    """TODO: Docstring for function.

    Args:
        i (int): TODO
        s (str): TODO
        o (MyObject): TODO

    Returns: TODO

    """
    pass
```


lpil (approved):
Thank you

PR #1356, opened 2021-07-09
sh: Add snippets for 'if' and 'func'

iiey:
- add portable if test expr
- add snippets to create function


lpil:
Could we have a different trigger for the new one? So it can be activated without having to pick which one. Thanks

iiey:
Sure, thank you for your review!
How is it now? Or, do you have any better suggestion?

lpil (approved):
Thank you

PR #1357, opened 2021-07-16
Fix syntax error: unexpected end of file

sharils:
By placing the closing TOKEN at the begin of line, this fixes `syntax error: unexpected end of file`.

`echo \\n` creates a mere newline without indentation of the current line unlike a new line in a snippet.

See https://github.com/koalaman/shellcheck/wiki/SC1039


lpil (approved):
Thank you

PR #1360, opened 2021-08-07
Fix wrong snippet commands

ujkuo:
I modify snippets for Integral and Partial derivatives for some typos.
E.g., lack of \ and the argument number.

Thanks for contributing snippets!


honza:
Thanks!

PR #1365, opened 2021-09-05
[Python] Handle type hints having ","

fwpp:
Enhance argument parsing for type hints having ","
, such as "arg: Tuple[int, int]".


lpil (approved):
Thank you

PR #1366, opened 2021-09-06
Adjust: use python3 default, same as snippets/python.snippets

jetz:
python2 is end of life, so make python3 as default, same as snippets/python.snippets does.

### DIFF
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -5,8 +5,8 @@ priority -50
 ###########################################################################
 
 #! header
-snippet #! "#!/usr/bin/env python3" b
-#!/usr/bin/env python3
+snippet #! "#!/usr/bin/env python" b
+#!/usr/bin/env python
 $0
 endsnippet
 
--- a/UltiSnips/python.snippets
+++ b/UltiSnips/python.snippets
@@ -5,8 +5,19 @@ priority -50
 ###########################################################################
 
 #! header
-snippet #! "#!/usr/bin/env python" b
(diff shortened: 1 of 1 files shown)


lpil:
As python2 is gone shouldn't `python` be python3?

jetz:
> As python2 is gone shouldn't `python` be python3?

I also hope `python`  referrs to python3, but See [PEP394](https://www.python.org/dev/peps/pep-0394). What `python` refers to differs depends on OS distributions or versions. So `python2` and `python3` is more clear, until pep394 is revised some day.

UltiSnips/python.snippets
```diff
@@ -5,8 +5,19 @@ priority -50
 ###########################################################################
 
 #! header
-snippet #! "#!/usr/bin/env python" b
-#!/usr/bin/env python
+snippet #! "#!/usr/bin/env python3" b
+#!/usr/bin/env python3
```
lpil, line 9: Can you make this one python with no number? The `#!3` snippet can be used for this.

jetz: `#!` with no number is shortcut for using python3, because python3 is used more frequently now，just like https://github.com/honza/vim-snippets/blob/master/snippets/python.snippets.

lpil: Yes but we wish to distinguish between the snippets rather than sharing triggers. `#!/usr/bin/env python` will run python3 on any up-to-date OS now.

lpil (approved):
Thank you 💜

PR #1367, opened 2021-09-09
[Python] Improve snippet tryef

fwpp:
Append space directly to except instead of using placeholder. For
providing space from placeholder, user need type space again when
modifying Exception class.


fwpp:
Sorry, I was wrong. The original version is perfect.

PR #1369, opened 2021-09-14
starting with snippets in react native

RaveHunter05:


lpil:
Could you move these out of the JavaScript directory? We do not want library snippets loaded by default. Thank you

PR #1370, opened 2021-09-17
Replace ancient form of python props

sedrubal:
I'm not sure if this is form of python properties are from python2 ages or even older.


lpil (approved):
Thanks!

PR #1373, opened 2021-10-07
feat(go): update error wrap snippet

b4nst:
Since go1.13 it's [now possible](https://go.dev/blog/go1.13-errors) to use `%w` verb to wrap an error natively. I think it makes more sense to have a native snippet than one that requires [an external library](https://pkg.go.dev/github.com/pkg/errors) to works


lpil (approved):
Thank you

PR #1379, opened 2021-10-31
feat(go): functional error handling

crackcomm:
Very convenient snippet, familiar to all VS Code users.

NOTE: It puts cursor before err, so one can conveniently `D` to remove `err` or just insert more return parameters.

### DIFF
--- a/UltiSnips/go.snippets
+++ b/UltiSnips/go.snippets
@@ -113,3 +113,9 @@ if err != nil {
 	log.${1:Fatal}(err)
 }
 endsnippet
+
+snippet iferr "Functional error handling" b
+if err != nil {
+	return ${1}err
+}
+endsnippet


UltiSnips/go.snippets
```diff
@@ -113,3 +113,9 @@ if err != nil {
 	log.${1:Fatal}(err)
 }
 endsnippet
+
+snippet iferr "Functional error handling" b
+if err != nil {
+	return ${1}err
```
lpil, line 119:
```suggestion
	return ${1:err}
```

How about this?

lpil:
Seeing as this doesn't use any ultisnips features it would be best to move this in the snippets directory so all plugins can use it.

crackcomm:
Oh, I didn't know I can use snippets from `snippets/` when using ultisnips. Looks like there is a [similar](https://github.com/honza/vim-snippets/blob/master/snippets/go.snippets#L84-L88) snippet in there already.

PR #1380, opened 2021-11-01
CSS: reorder `text-shadow`'s arguments and add `inset` to `box-shadow`

Aster89:
[`box-shadow`](https://www.w3schools.com/CSSref/css3_pr_box-shadow.asp) and [`text-shadow`](https://www.w3schools.com/cssref/css3_pr_text-shadow.asp) have a very similar syntax, so it makes sense to have the order of the arguments common to the two properties be the same, so I moved `color` to the end for `text-shadow` to match the order of `box-shadow`.

I've also added `inset` to `box-shadow`, because why not.

As regards the opening of the snippets, should I reorder also that one? I'm asking because the order of the arguments on lines 51 and 52 wasn't the same even before my change. By the way, what does the `(text)` thing mean?

### DIFF
--- a/UltiSnips/css.snippets
+++ b/UltiSnips/css.snippets
@@ -45,11 +45,11 @@ snippet ! "!important CSS (!)"
 endsnippet
 
 snippet tsh "text-shadow: color-hex x y blur (text)"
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1387, opened 2021-12-26
Replace f-string with format

firstmover:
f-string only works for python3.6 or above. The snippet fails to work when the vim specified python version is below 3.6.

[Related issue](https://github.com/neoclide/coc-snippets/issues/214)


lpil:
Python 3.5 and Python 3.6 have both been end of lifed and no longer get any security updates. Please urgently upgrade to a supported version of Python.

PR #1388, opened 2021-12-28
add ruby3 debug snippet

r3jack:
https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/
https://github.com/ruby/debug


lpil (approved):
Thank you

PR #1390, opened 2022-01-11
Update snippets

SingularisArt:
Hello all.

This is one of my first pull requests. So, tell me if I'm doing anything wrong because I won't know.
Anywho, I updated the latex, lua, and gitcommit snippets.

For latex, I added a trillion more snippets and updated the current snippets.
For lua, I just added two more snippets. One for require and one for use.
For gitcommit, I removed all snippets and created 3 snippets.
    Instead of having different snippets like docs, refactor, test, etc I created a snippet that is based on python and while you type, it will give you that result.
    I added a signature snippet that will give you your signature.
    I added a good commit structure with details on what to say and what not to say as well as information on how to word and structure your sentences.

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -1,66 +1,698 @@
-priority -50
-
-extends texmath
+##########
+# GLOBAL #
+##########
 
 global !p
 
 def create_table(snip):
 	rows = snip.buffer[snip.line].split('x')[0]
 	cols = snip.buffer[snip.line].split('x')[1]
-
 	int_val = lambda string: int(''.join(s for s in string if s.isdigit()))
-
 	rows = int_val(rows)
 	cols = int_val(cols)
-
 	offset = cols + 1
 	old_spacing = snip.buffer[snip.line][:snip.buffer[snip.line].rfind('\t') + 1]
-
 	snip.buffer[snip.line] = ''
-
 	final_str = old_spacing + "\\begin{tabular}{|" + "|".join(['$' + str(i + 1) for i in range(cols)]) + "|}\n"
-
 	for i in range(rows):
 		final_str += old_spacing + '\t'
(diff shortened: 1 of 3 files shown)


lpil:
Hello! Would you mind putting back the delete snippets? Someone may be using them.

SingularisArt:
Alright. I added the snippets that I removed from the gitcommit.snippets file.
Quick question. How do I add that commit to this pull request?

lpil (approved):
Thank you

lygamac:
@SingularisArt can you remove the auto trigger tag for `use` snippet in `lua`? It gets triggered when typing common words like `user`

SingularisArt:
Ah. Yes. I'll do that.

UltiSnips/tex.snippets
```diff
@@ -1,258 +1,871 @@
...
+\chapter{$1${VISUAL}}
 endsnippet
 
-snippet lp "Long parenthesis"
-\left(${1:${VISUAL:contents}}\right)$0
+snippet sec "Section" wi
+\section{$1${VISUAL}}
 endsnippet
 
-snippet "mint(ed)?( (\S+))?" "Minted code typeset" br
```
sanjayankur31, line 209: Hi folks: was this an intentional removal?

SingularisArt:
@sanjayankur31 Sorry for the headache I've caused. I'll keep that in mind when doing another PR.

sanjayankur31:
No worries.

So was dropping the minted snippet intentional or was that just a fall out from the refactor? Were there other snippets that were dropped here, and were they all intentional?

I personally use minted quite a bit so I'd like to have the snippet, but I can keep it in a personal repo if folks think it's too specific to be here. If it should be here, I'll open a PR to add it back.

ThomasFeher:
Why was the labeling for the section commands removed?

Issue #1391, opened 2022-01-14
Tex: New snippet for fraction is too intrusive


cristobaltapia:
The snippet defined [here](https://github.com/honza/vim-snippets/blob/master/UltiSnips/tex.snippets#L352#L354) (see below) inserts a fraction (`\frac{ }{ }`) every time that the `/` (slash) is used. This is a symbol that is commonly used also outside of the mathematical scope and should not autotrigger without asking (e.g. when writing a file path). If it is OK for you, I would propose to either remove it or make it aware of the math context before it autotriggers.

```snip
snippet / "Fraction" i
\\frac{${VISUAL}}{$1}$0
endsnippet
```

lpil:
Let's remove it. Would you like to make a PR @cristobaltapia ?

cristobaltapia:
Already done :)

mowgli:
Sorry, Should have done a pull before.

Thanks

PR #1392, opened 2022-01-14
fix: remove snippet that autotriggers \frac with /

cristobaltapia:
This snippet was removed since it autotriggers regardless of whether the
cursor is in a math environment or not.


cristobaltapia:
fixes #1391

lpil (approved):
Thank you

PR #1393, opened 2022-01-14
Inline Math and Display Math Ultisnippet fixed

MatteoGaetzner:
The previous implementation would cause a python exception because of wrong array indexing.
The proposed implementation gets rid of the python stuff, which I don't really understand is the purpose of in the first place, which fixes the error.

Additionally, the snippets 'im' and 'dm' now don't get triggered automatically anymore, which can be annoying if you want to write text with prefix 'im' or 'dm', for example 'image' and 'dmesg'.

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -322,16 +322,11 @@ snippet DD "D" w
 \mathbb{D}
 endsnippet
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1394, opened 2022-01-14
fix(tex): remove problematic snippets added recently

cristobaltapia:
General comment:

There are too many snippets that are too specific. I don't think that they should all be here. That is what personal repos are for. Also, too many autotriggers, especially for math operations without checking that we are in a math environment.

I would propose to remove/fix the following snippets:

- \frac snippet (autotrigger)
- \times with * (we need this for multiplications e.g. in tikz)
- \ldots autotrigger with "..." (this also is to intrussive as is needed in tikz)
- Autosubscripts (It should be possible to write T9 withouth getting a subscript. All this snippets are way too specific for one person's use)
- Superscript (redundant snippets. One suffices as includes all the others... and that also does not convince me, but we can leave it.)
- "to the power" (rd) (another autotrigger that is annoying. In German you write approx. as "rd.")
- Subscript is sensible
- "->" (if this should be mapped to something, then to an arrow. But I would preffer to leave it out.)
- EE ... again an autotrigger we should not have
- AA ... another autotrigger that will annoy people at some point
- xp1 (super specific)
- R0+ (too specific)
- nabla (should not autotrigger)
- norm (another autotrigger that annoys)
- notin (so we can write 'noting' again)
- <> (we don't need a snippet for \diamond)
- text subscript trigger by '(?<!i)sts' (still triggers with 'toasts')
- set (also not a nice autotrigger to have)
- \\\ (don't mess with my escape sequences)
- >> and << might conflict with tikz'a arrows
- <-> also is a problem with tikz
- dint (also an autotrigger that bothers me)
- arcsin|arccos|arctan|arccot|arccsc|arcsec|pi|zeta|int (also not)
- The 'bar' snippets (another autotrigger)
- 'doc' (common, you cannot make this autotrigger... how many times do you even use this?)
- beg should not autotrigger (we can leave it though)
- cnt also should not autotrigger
- desc should not autotrigger
- lemma should not autotrigger
- prop should not autotrigger
- thrm should not autotrigger
- post should not autotrigger
- prf should not autotrigger
- def should not autotrigger (ever!)
- nte should not autotrigger
- prob should not autotrigger
- corl should not autotrigger
- notion should not autotrigger
- rep should not autotrigger
- prop (again!) should not autotrigger
- int should not autotrigger
- obs should not autotrigger
- conc should not autotrigger
- fig should not autotrigger (!!!!!)
- enum should not autotrigger
- item should not autotrigger
- case should not autotrigger
- ali should not autotrigger
- ali* should not autotrigger
- box (why do we need this for a latex document? This should go to general snippets)
- vector postfix is too specific and autotriggers
- date snippet is also a general snippet (and already exists)
- signature (we are not you)
- ref should not be linked to referencing figures as it is used to reference all kinds of things.
- lesson (too specific)
- lecture (too specific)
- newpage (autotrigger that is not really needed)

There are certainly other snippets that I would delete, but these are the
ones that I consider most problematic.

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -36,79 +36,73 @@ endglobal
 # ENVIRONMENT #
 ###############
 
-snippet beg "begin{} / end{}" bAi
+snippet beg "begin{} / end{}" bi
 \begin{$1}
 	$0${VISUAL}
 \end{$1}
 endsnippet
 
-snippet doc "Document" bAi
-\begin{document}
-	$0${VISUAL}
-\end{document}
-endsnippet
-
-snippet cnt "Center" bAi
+snippet cnt "Center" bi
 \begin{center}
 	$0${VISUAL}
 \end{center}
 endsnippet
 
-snippet desc "Description" bAi
+snippet desc "Description" bi
 \begin{description}
 	$0${VISUAL}
 \end{description}
 endsnippet
 
-snippet lemma "Lemma" bAi
+snippet lemma "Lemma" bi
 \begin{lemma}
 	$0${VISUAL}
 \end{lemma}
 endsnippet
 
-snippet prop "Proposition" bAi
+snippet prop "Proposition" bi
 \begin{prop}[$1]
 	$0${VISUAL}
 \end{prop}
 endsnippet
 
-snippet thrm "Theorem" bAi
+snippet thrm "Theorem" bi
 \begin{theorem}[$1]
 	$0${VISUAL}
 \end{theorem}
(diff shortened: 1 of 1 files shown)


cristobaltapia:
I would almost rather revert the original commit.

lpil (approved):
Thanks for going through these and sorting them

cristobaltapia:
Sure, no problem. I use latex snippets almost daily so I want them to work :). I will review these again later and suggests changes if needed. Have a nice weekend!

PR #1395, opened 2022-01-15
cleanup(tex): improvements and cleanup for tex snippets

cristobaltapia:
Hi, its me again with some TeX improvements/corrections. :)

Summary of all commits:

f720db1 fix(tex): unclear use of `${VISUAL}` here
1563422 fix(tex): correct use of `${VISUAL}` placeholder
3c10e8a cleanup(tex): a more sensible snippet for `description`
3390ab4 cleanup(tex): the environment `intuition` is not standard
df5e78c cleanup(tex): the environment `property` is not standard
598cc6b cleanup(tex): the environment `repetition` is not standard
8b37206 cleanup(tex): the environment observation is not standard
8fba2b6 cleanup(tex): align is already defined in texmath
0a53561 cleanup(tex): these are all also rather specific snippets
a4bac70 cleanup(tex): `\N` is not a standard latex command
3a2f74f cleanup(tex): these snippets were too specific
fdcbb70 fix(tex): inline math should use `\(...\)` instead of `$...$`.
156f9d8 cleanup(tex): more `\frac{}{}` that autotrigger when not needed
b68af00 cleanup(tex): subscript and superscript would trigger outside math
5c9c355 cleanup(tex): plot snippet is too specific.
2441209 fix(tex): SI snippet is already in texmath

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -138,12 +138,6 @@ snippet int "Intuition" bi
 \end{intuition}
 endsnippet
 
-snippet obs "Observation" bi
-\begin{observation}[$1]
-	$0${VISUAL}
-\end{observation}
-endsnippet
-
 snippet conc "Conclusion" bi
 \begin{conclusion}[$1]
 	$0${VISUAL}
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1397, opened 2022-01-22
Remove too specific tex snippets

smjonas:
I find these pretty annoying so I made the following changes
- removed `$$` snippet (I think this was the most annoying one, especially with autotrigger when you wanted to write math like `$$...$$` in latex)
- changed `//` trigger to `frac`
- removed `~~` snippet
- removed `||` snippet
- removed `**` snippet

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -213,17 +213,13 @@ snippet DD "D" w
 \mathbb{D}
 endsnippet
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks

cxr6548:
Can the inline math be re-added with another trigger? I use it a lot, even more than display math. Perhaps something like in$?

PR #1401, opened 2022-02-02
feat(css): add web component APIs

bennypowers:
- adds [`:host`](https://developer.mozilla.org/en-US/docs/Web/CSS/:host()) and `:host()`
- adds [CSS Shadow Parts](https://www.w3.org/TR/css-shadow-parts-1/)
- adds [CSS Custom Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

### DIFF
--- a/snippets/css.snippets
+++ b/snippets/css.snippets
@@ -1245,3 +1245,19 @@ snippet ::a
 	::after
 snippet ::b
 	::before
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #1402, opened 2022-02-22
Update README.md and latex snippets

SingularisArt:
Hello.

I updated the latex snippets.
I also updated the README.md for the UltiSnips directory. Now, it contains all of the information for the current snippets.

It goes through each file type and talks about their snippets that they have.


lpil (commented):
This is a very large contribution! I don't think I have the ability to evaluate its suitability as I'm not a tex user. Last time we merged a large change like this it caused a lot of issues for people's existing workflows

SingularisArt:
Would you like me to break the commit into multiple ones like first modifying the README, then each snippet file?

PR #1406, opened 2022-03-13
Fix formatting error

0q2:
Was wondering why I was getting an error from UltiSnips. Once I looked at the most recent merge it was pretty obvious.


lpil (approved):
Oops, thank you

PR #1408, opened 2022-03-18
Remove autotrigger of frac

jimmy-zx:
The recent PR #1397 has changed ``//`` trigger to ``frac``, but it remains to be autotriggered. Each time I wrote ``\frac`` it would be automatically expanded to ``\\frac{}{}``. This behavior would be better if it could be triggered manually.


lpil (approved):
Thank you

PR #1412, opened 2022-03-26
Autotrigger behavior of bra and ket

jimmy-zx:
When writing comparisons with absolute values, like

formula

Typing `|a|>` the snippet would autotrigger and convert it into `\ket{a|}`. There were also similar behavior with `\ker`.

Seems like these snippets are related to physics, but are disturbing when typing math formulas.


lpil (approved):
Thank you

PR #1413, opened 2022-03-27
Remove outdated Rails `sweeper` snippet

andyw8:
When browsing the Rails snippets, I noticed this snippet for a feature that was removed in Rails 3.2.13, over nine years ago:

https://apidock.com/rails/ActionController/Caching/Sweeper


lpil (approved):
Thank you

PR #1414, opened 2022-03-27
Clarify README re. `Ultisnips`

andyw8:
I'm new to this project, but I _think_ this is what was intended.


lpil:
It was right before, the ultisnips snippets have been placed into the ultisnips directory. Going forward we want to have as few as them as possible as they cannot be read by all snippet plugins.

andyw8:
Oh, I see. I've pushed an update that hopefully makes that clearer.

lpil (approved):
Thank you

PR #1416, opened 2022-03-28
Re-add Inline Math snippet

cxr6548:
mimics the function of Display Math with a less obtrusive trigger than before


lpil (approved):
Thanks

PR #1417, opened 2022-04-01
main snippet for Kotlin

flwyd:
`main` matches the name for other language snippets for the entry point to a program.


lpil (approved):
Thank you

PR #1420, opened 2022-04-08
added basic flutter scaffold application snippet

andromedabushturkey:
As the title states, I've added a basic flutter scaffold application with the trigger of 'fsa'(flutter scaffold application)


lpil (approved):
Thank you

PR #1424, opened 2022-04-27
remove bracket of definition part of scheme.snippet `def`

gzqx:
Since `if` `cond` and other snippets all include brackets around them, including brackets in `def` is duplicate and will lead to an error if not deleted (e.g. if `if` returns a value), IMHO not including brackets is more intuitive.


lpil (approved):
Thank you

PR #1428, opened 2022-05-08
Added basic snippets for Linux asm x86

fredtux:
Hi! I've added some snippets that I use for assembly x86. Maybe the community is interested as well


lpil (commented):
It looks like these snippets don't use any ultisnips features so they would be best in the snipmate format. Would you mind coverting them over? Thanks

smjonas:
I could do the conversion part in a new PR if you'd like me to (using the snippet conversion plugin I created). Are there any existing snippets that should be converted from or to UltiSnips / SnipMate?

lpil:
There's probably lots but I could not tell you offhand which ones they are

fredtux:
I've converted them to snipmate format. Hope it's ok this time

lpil (approved):
Wonderful, thank you

PR #1429, opened 2022-05-09
Add basic snippets for Linux asm x86

smjonas:
I converted the snippets provided by fredtux to SnipMate format (see #1428). 
@fredtux: feel free to copy these into your PR! (this PR is not intended to be merged).

### DIFF
--- a/snippets/asm.snippets
+++ b/snippets/asm.snippets
@@ -1,0 +1,212 @@
+snippet scode Start basic code for assembly
+	.data
+
+
+	.text
+
+
+	.global main
+
+
+	main:
+
+
+snippet scodes Start basic code for assembly with _start label
+	.data
+
+
+	.text
+
+
+	.globl _start
+
+
+	_start:
+
+
+snippet lo Long
+	$1: .long $2
+snippet wo Word
+	$1: .word $2
+snippet by Byte
+	$1: .byte $2
+snippet sp Space
+	$1: .space $2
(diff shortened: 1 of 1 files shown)


fredtux:
> I just noticed some comments that have not been handled correctly and I'll work on that :D (comments cannot be included inline, only at the start of a snippet definition). I still think this is a good starting point.

Sorry, I can't figure out which ones they are, please give me an example and I'll change the rest. Thanks :)

smjonas:
Sure, e.g. in line 53 there is:
```snippets
snippet exit0 Exit without error
	# Exit without error
	movl \$1, %eax
	xorl %ebx, %ebx
	int \$0x80
```
while it should be 
```snippets
# Exit without error
snippet exit0 Exit without error
	movl \$1, %eax
	xorl %ebx, %ebx
	int \$0x80
```
for all snippets.
For context, this is an extract from the SnipMate docs:
```
Comments can be made in .snippets files by starting a line with a # character.
However these can't be used inside of snippet definitions: >
    # this is a correct comment
    snippet trigger
        expanded text
    snippet another_trigger
        # this isn't a comment!
        expanded text
```

fredtux:
Right, I used those inside assembly to have a reference for what the code was doing since it's very obscure. I see I can't escape the character using \ . Is there any way to add comments as part of the snippet? If not, I will simply remove them.

PR #1431, opened 2022-05-15
Markdown table fix and enhancement

pilgrimlyieu:
1. Fix markdown table unexpected overflow bug
2. Add different types of table support

Since previous markdown table version has only left-aligned type, I add different types support. However, these changes cause unexpected overflow bug, which appears only in the center-aligned type, and I couldn't find out where it lies in. So I transform `post_jump` into `pre_expand` according to SirVer/ultisnips/issues/877.

### DIFF
--- a/UltiSnips/markdown.snippets
+++ b/UltiSnips/markdown.snippets
@@ -1,18 +1,18 @@
 priority -50
 
 global !p
+
+table_settings = [["-", "-"], [":", "-"], ["-", ":"], [":", ":"]]
+
 # A overkill(dirty) table with automatic alignment feature
-def create_table(snip):
+def create_table(snip, align = 0):
 	# retrieving single line from current string and treat it like tabstops count
-	placeholders_string = snip.buffer[snip.line].strip()
+	placeholders_string = snip.buffer[snip.line].strip()[-2:]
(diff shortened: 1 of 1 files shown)


pilgrimlyieu:
One thing should be noticed is that you can only use this snippet in an empty line, which means

```markdown
tb22aaaa
<!-- You first enter "aaaa" and then use "tb22" to trigger the snippet. -->
```

will throw error.

lpil:
Hi @pilgrimlyieu , I think we would not want it to throw an error?

pilgrimlyieu:
> Hi @pilgrimlyieu , I think we would not want it to throw an error?

That's true, However, Markdown table always should be created in an empty line or it will not take effect.

pilgrimlyieu:
It will not throw error while triggering table snippets in an unempty line any more.

pilgrimlyieu:
It will preserve text after trigger word now.

```markdown
tb33<tab>abc
```

will now be expanded as

```markdown
| head1 | head2 | head3 | 
|-------|-------|-------|
| R1C1  | R1C2  | R1C3  | 
| R2C1  | R2C2  | R2C3  | 
| R3C1  | R3C2  | R3C3  | 
abc
```

PR #1432, opened 2022-05-25
Add bar to snippets/tex.snippets

kiryph:
exists already in UltiSnips/tex.snippets

### DIFF
--- a/snippets/tex.snippets
+++ b/snippets/tex.snippets
@@ -406,6 +406,10 @@ snippet ddot ddot
 snippet vec vector
 	\\vec{{$1}} {$0}
 
+# Bar
+snippet bar bar
+	\\bar{{$1}} {$0}
+
 # Cross product
 snippet \x cross product
 	\\times {$0}


kiryph:
> Could you delete it from the ultisnips dir please

Do UltiSnips user loose this snippet then or does UltiSnips by default load both?

kiryph:
I think the UltiSnips snippet is smarter: the bar snippet is categorised under POSTFIX with the flags `wi`
```
   i   In-word expansion - By default a snippet is expanded only if the tab
       trigger is the first word on the line or is preceded by one or more
       whitespace characters. A snippet with this option is expanded
       regardless of the preceding character. In other words, the snippet can
       be triggered in the middle of a word.
   w   Word boundary - With this option, the snippet is expanded if
       the tab trigger start matches a word boundary and the tab trigger end
       matches a word boundary. In other words the tab trigger must be
       preceded and followed by non-word characters. Word characters are
       defined by the 'iskeyword' setting. Use this option, for example, to
       permit expansion where the tab trigger follows punctuation without
       expanding suffixes of larger words. This option overrides 'i'.
```
and support for visual selection.

```
###########
# POSTFIX #
###########

snippet bar "bar" wi
\bar{${1:${VISUAL}}}$0
endsnippet
```

I have used UltiSnips before and if I remember it correctly, UltiSnips will only consider one bar snippet as long as they are both named `bar` and the UltiSnips snippet has higher priority.

lpil:
Ok, let's keep it then. Thank you

PR #1435, opened 2022-06-10
correct $ as inline math delimiter fault in tex.snippets

sqliuchang:
I'm so sorry for my fault before.


sqliuchang:
If $ is not necessary for most people, please help me delete these error codes.
I'm really sorry for all this trouble.

PR #1439, opened 2022-08-14
Convert the UltiSnips tex snippet `tmplt` to snipmate format

kiryph:
The snippet does not use any features specific to UltiSnips.
Now it is also available for Snipmate, LuaSnip, ... user.


lpil (approved):
Thank you

PR #1440, opened 2022-08-14
Update lua.snippets

SingularisArt:
Remove the auto-trigger for the snippets `use` in the lua filetype. [This](https://github.com/honza/vim-snippets/pull/1390#issuecomment-1214369632) comment brought that to my attention.


lpil (approved):
Thank you

PR #1441, opened 2022-08-16
Move cursor to the next line after include in C

amarakon:
With the snippet `inc` in `c.snippets`, I think it is better to have the caret at the next line after pressing `Tab` rather than at the end of the line.

```vim
inc<Tab><Tab>
```

```c
#include <stdio.h>
|
```


lpil (approved):
Thanks

PR #1443, opened 2022-08-23
Add `print` function snippets for Python

amarakon:
I noticed that the `python.snippets` file does not have any `print` function snippets, so I decided to add some similar to the `printf` snippets from `c.snippets`.


lpil (approved):
Thank you

PR #1444, opened 2022-08-25
Improve print and write functions for Lua

amarakon:
So far `lua.snippets` just has the `print` snippet. I decided to add a couple more snippets for printing, including one with `string.format` to format the string. I also added snippets for printing with `io.write` (to omit the trailing newline) and also added snippets with writing to `stderr`.


lpil (approved):
Thanks!

PR #1446, opened 2022-08-27
Redefine `mainn` snippet for C++ without `void`

amarakon:
In C++ (not C), `int main(void)` and `int main()` do the exact same thing but it is conventional to write `int main()` instead because it is shorter.


lpil (approved):
Thank you

PR #1447, opened 2022-08-27
Improve POSIX-compliant shell snippets

amarakon:
Here are the changes I made to `sh.snippets`, `zsh.snippets` and `bash.snippets` (the latter I created the file).

- remove Bashisms from `sh.snippets`
  - Bashisms are things that can't be run in shells other than Bash, so we should use pure POSIX-compliant code in `sh.snippets` and put Bashims in `bash.snippets` instead
- create the file `bash.snippets` which extends `sh` but uses the Bashims
- change the `#!` snippet to `#!/usr/bin/env bash` instead of `#!/usr/bin/env sh` in `bash.snippets`
- do the same above for `zsh.snippets`
- make `zsh.snippets` extend `bash`

### DIFF
--- a/snippets/bash.snippets
+++ b/snippets/bash.snippets
@@ -1,0 +1,25 @@
+extends sh
+
+# Shebang
+snippet #!
+	#!/usr/bin/env bash
+
+snippet s#!
+	#!/usr/bin/env bash
+	set -eu
+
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you

PR #1449, opened 2022-08-29
Add nvim-snippy and LuaSnip to snippet engines

amarakon:
I added nvim-snippy and LuaSnip to the snippet engines section in the README for people who use Neovim and prefer Lua plugins.


lpil (approved):
Thank you

PR #1452, opened 2022-08-30
Support `while true` loops

amarakon:
For `while` loops in several languages, the default condition is true. Snippets called `wht` are also added in which the condition is true. I did this because having the default condition as true is more useful than having a comment there. `while true` is used commonly for making a forever loop.

### DIFF
--- a/snippets/rust.snippets
+++ b/snippets/rust.snippets
@@ -147,7 +147,11 @@ snippet loop "loop {}" b
 		${0:${VISUAL}}
 	}
 snippet wh "while loop"
-	while ${1:condition} {
+	while ${1:true} {
+		${0:${VISUAL}}
+	}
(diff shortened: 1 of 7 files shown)


snippets/rust.snippets
```diff
@@ -147,7 +147,11 @@ snippet loop "loop {}" b
 		${0:${VISUAL}}
 	}
 snippet wh "while loop"
-	while ${1:condition} {
+	while ${1:true} {
```
lpil, line 150: This is not the idiomatic way to make a loop with no conditional in rust. could you revert this one please? Thanks

amarakon: It is fixed in the latest commit.

lpil (approved):
Thank you!

PR #1453, opened 2022-09-01
Improve Markdown and R Markdown snippets

amarakon:
Here are the changes I made to improve Markdown and R Markdown snippets.

- use snippet descriptions instead of unneeded placeholder text
- add a font matter snippet which is very useful for Pandoc
- add another definition list snippet
- since R Markdown is just Markdown with extra things, add `extends markdown` to `rmd.snippets` and remove duplicate snippets
- delete the extra blank line in `pandoc.snippets`

### DIFF
--- a/snippets/markdown.snippets
+++ b/snippets/markdown.snippets
@@ -67,17 +67,21 @@ snippet <*
 	<`@*`>
 snippet <c
 	<`@+`>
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you

PR #1455, opened 2022-09-04
add snippets for smarty

shadowwa:
Hi,

Here is a snippets file for smarty template engine (https://www.smarty.net/) a php template engine.
It's usualy used to generate _html_, _css_ or _javascript_ files, hence the "extends" on these three languages in this file.

### DIFF
--- a/UltiSnips/smarty.snippets
+++ b/UltiSnips/smarty.snippets
@@ -4,81 +4,6 @@ extends html
 extends javascript
 extends css
 
-# https://www.smarty.net/docs/en/language.function.if.tpl
-snippet if "{if cond} ... {/if}"
-{if ${1}}
-	${0:${VISUAL}}
-{/if}
-endsnippet
-
-snippet ifn "{if !cond} ... {/if}"
-{if ${1}}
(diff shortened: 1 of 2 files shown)


shadowwa:
I just figured out that coc-snippets support a notation for a choice selection as a placeholder.
for example 
```
scope='${1|parent,root,global|}'
```
instread of
```
scope='${1:parent|root|global}'
```
but I did not found this kind of notation in vim-snippets. 

is this notation to give a choice selection standard enough to be include in vim-snippets or is it specific to the coc-snippets implementation?

smjonas:
Your first notation for choice snippets is only supported by UltiSnips, the original SnipMate syntax does not support it. So that snippet should probably be in the UltiSnips directory.

lpil (commented):
Could you convert these to the snipmate format please? That is the preferred format unless ultisnips specific features are being used.

smjonas:
Pretty sure he intends to say that you can simply undo your last commit and change the snippets with choice selection to UltiSnips, i.e. your option 2.

lpil (approved):
Thank you

PR #1457, opened 2022-09-05
Remove unnecessary placeholder text

amarakon:
This [section](https://github.com/honza/vim-snippets#things-to-consider-when-contributing) of the README states:

> Don't add useless placeholder default texts

A lot of snippets still do this though, so I used a regular expression and some manual editing to fix it.


lpil (approved):
Thank you!

PR #1458, opened 2022-09-07
Remove unnecessary body placeholder text

amarakon:
Just like the last commit, I removed more unnecessary placeholder text.


lpil (approved):
Thank you

PR #1459, opened 2022-09-11
Snippet for #pragma once include guard

Aster89:
It is non-standard, but most (all?) compilers implement it.

See also https://en.wikipedia.org/wiki/Pragma_once#Portability


lpil (approved):
Thanks

PR #1460, opened 2022-09-11
Snippets for Range-v3 and C++20's <ranges> entities

Aster89:
I find myself very frequently writing `ranges::views::`, and I'd be doing the same with `std::views::` if I was using C++20, so I think this snippets can be useful.

`transform` and `filter` are two of the most used adapters, so haveing a snippet for them is good (for both `std::` and `ranges::`, at the price of a backspace).

### DIFF
--- a/snippets/cpp.snippets
+++ b/snippets/cpp.snippets
@@ -230,3 +230,24 @@ snippet af auto function
 	{
 		${0}
 	};
+# Range-v3 transform
+snippet transform "ranges::views::transform"
+	${1:${2:std::}${3:ranges::}views::}transform($4)
+# Range-v3 transform
+snippet filter "ranges::views::filter"
+	${1:${2:std::}${3:ranges::}views::}filter($4)
+# Range-v3 ranges::
+snippet r "ranges::"
+	ranges::
+# Range-v3 ranges::views::
+snippet rv "ranges::views::"
+	ranges::views::
+# Range-v3 ranges::actions::
+snippet ra "ranges::actions::"
+	ranges::actions::
+# STL std::ranges::
+snippet sr "std::ranges::"
+	std::ranges::
+# STL std::views::
+snippet sv "std::views::"
+	std::views::


Aster89:
I think this should be reviewed by some more knowledgeable than me snippet author.

The problem is that sombody (well, me) sometimes uses to write `namespace rv = ranges::views;` at the top of cpp files.

In case I'm in the first tab-stop of [this snippet](https://github.com/honza/vim-snippets/pull/1461/files#diff-03bab7f0c5baaa691365cfcbe5fd2a5977497384991884606d6b978d273a3462R75-R77), hitting <kbd>F8</kbd> (or equivalent) triggers the expansion of this.

Anybody knows how this can be avoided?

Or should I just change the trigger to something longer than two letters?

smjonas:
Which snippet engine are you using? At least for `vim-snipmate`, pressing `<Tab>` won't try to expand the inner `rv` snippet if there is a next tabstop by default so this shouldn't be a problem. From the SnipMate [docs](https://github.com/garbas/vim-snipmate/blob/525f331320427bf7aeb07651e2536b1f8a23df03/doc/SnipMate.txt#L198):
> \<Plug\>snipMateNextOrTrigger 
> Default: <Tab\> 
> Jumps to the next tab stop or, **if none exists**, try to expand a snippet.

Aster89:
Makes sense. I use UltiSnips, and apparently I've got this in my vimrc,
```viml
let g:UltiSnipsExpandTrigger = "<F8>"
let g:UltiSnipsJumpForwardTrigger = "<F8>"
```
which explains the behavior I had got used to.

I guess you can merge this PR as is, if you like it.

UltiSnips/cpp.snippets
```diff
@@ -196,5 +196,33 @@ BOOST_AUTO_TEST_CASE(${4:TestCaseName})
...
+snippet ra "ranges::actions::"
+ranges::actions::
+endsnippet
+
+snippet sr "std::ranges::"
+std::ranges::
+endsnippet
+
+snippet sv "std::views::"
+std::views::
```
lpil, line 226: These don't use any ultisnips features, could you add them using the snipmate format instead? Thanks

Aster89: You mean moving them [in the other directory](https://github.com/honza/vim-snippets/tree/master/snippets) and using the format I find there?

lpil: Yup!

lpil (approved):
Thank you!

PR #1461, opened 2022-09-11
C++ snippet for namespace aliases and for using directives/declarations

Aster89:
I think this is a fairly usabel snippet:

[asciicast](https://asciinema.org/a/BgFg6ZSfLVyrnmlE8fPXoSaZp)


lpil (approved):
Thank you!

PR #1465, opened 2022-10-10
Improvements to the GenServer and Supervisor snippets

craigp:


snippets/elixir.snippets
```diff
@@ -265,22 +265,27 @@ snippet genserver basic genserver structure
...
-		GenServer.start_link(__MODULE__, ${1:Map.new})
+	def start_link(init_args) do
+		GenServer.start_link(__MODULE__, init_args, name: __MODULE__)
 	end
 
 	@impl true
-	def init(state) do
-		{:ok, state}
+	def init(_init_args) do
+		{:ok, []}
```
lpil, line 274: Let's keep with the previous pattern of passing in the state.

craigp: The reason I changed it is that although it might be state you're passing it, it might not be - generate it is some sort of initialisation info, which might include initial state.

lpil: Yes please

craigp: Done :)

lpil (approved):
Thank you

PR #1471, opened 2022-11-28
Disable auto trigger for leq and geq

jimmy-zx:
Disabled auto expansions of the following snippets
- `>=` -> `\ge`
- `<=` -> `\le`

When writing inline code using `\verb` or `minted`, the manual behavior would be more suitable.


lpil (approved):
Thank you

PR #1474, opened 2022-12-23
LaTeX snippets for sectioning commands include labels

Aster89:
Inspired by the snippet for the `figure` environment, I thought it could be useful to have snippets for sectioning commands that also come with a `\label` that converts all `\W` characters to `-`.

Personally, I'd also like to lowercase all letters, but I'm not sure if this would be welcome. I'm open to it though.


lpil (approved):
Thank you

PR #1475, opened 2023-01-13
Fixed flutter constructors and createState definition

guildem:
Default one-line constructors are missing and createState() definition needs to return State<T> object.
This pull-request fixes linting errors.


lpil (approved):
Thanks

PR #1478, opened 2023-01-23
Remove invasive autotrigger for LaTeX snippet (fixes #1476)

### LINKED ISSUE #1476: This lll snippets is very annoying when one is filling the second arg of a tabular environment.
Given

https://github.com/honza/vim-snippets/blob/1049d2425b6d270621b48e110de37d5467a2a621/UltiSnips/tex.snippets#L387-L389

I'm not able to write `llllll` in a `\begin{tabular}{ here }`. I think a better trigger should be used.

Aster89:


lpil (approved):
Thank you

PR #1479, opened 2023-01-23
Fix LaTeX snippet for inverse

Aster89:
If you type <kbd>1</kbd><kbd>2</kbd><kbd>3</kbd><kbd>i</kbd><kbd>n</kbd><kbd>v</kbd><kbd>s</kbd>, you'll see that it will be expanded to `3^{-1}`, throwing away anything but the last character before `invs`. This fixes it.

### DIFF
--- a/UltiSnips/tex.snippets
+++ b/UltiSnips/tex.snippets
@@ -400,7 +400,7 @@ snippet <! "normal" Aw
 \triangleleft 
 endsnippet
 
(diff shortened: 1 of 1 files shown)


Aster89:
Oh, by the way, does exist in Python regexes something like Vim's [`\%(`](https://vimhelp.org/pattern.txt.html#%2F%5C%25%28)?

PR #1480, opened 2023-01-24
Replace the hash rocket syntax from ruby with a colon

viktorsmari:
Is the snippet (UltiSnip/eruby) not being used anymore?

I noticed that the last change was in 2016, are ruby developers using a different snippet?

There are 2 snippets, I followed the readme setup and it ended up using the old one
 
**/snippets/eruby** - updated (2022)
- https://github.com/honza/vim-snippets/blob/master/snippets/eruby.snippets

**/UltiSnip/eruby** - updated in (2016)
- https://github.com/honza/vim-snippets/blob/master/UltiSnips/eruby.snippets

I have been using vim for 8 years and neovim for the last 4 years and have this setup:

```
Plug 'SirVer/ultisnips' 
Plug 'honza/vim-snippets'
```

And when I do a `rp` it produces the old syntax:
```ruby
render :partial => "item"
```

and `ft` :
```ruby
<% form_tag(:action => 'update', class => 'form'}) do %>
  
<% end %>
```

### DIFF
--- a/UltiSnips/eruby.snippets
+++ b/UltiSnips/eruby.snippets
@@ -44,13 +44,13 @@ snippet fi "<%= Fixtures.identify(:symbol) %>"
 endsnippet
 
 snippet ft "form_tag" w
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks!

Issue #1198, opened 2020-02-27
Triple-quote snippet for python produces unbalanced string


idbrii:
Using the `'` snippet doesn't insert the correct number of quotes. Instead, you get this:

```
'''
doc
'
```

Also occurs for `"` snippet.

# Repro

```vim
:edit test.py
:exe "normal a'\<Tab>"
```

Looks like it's due to `triple_quotes_handle_trailing` trying to handle autoclose plugins but I don't have one. Removing the `while` block fixes the issue.

# Version

SirVer/ultisnips@7dc30c55e5c41c98a8c7421bb01fec1d559256fd (3.2)
snippets 35067caeca1c17fa5d38d5de031b26610e939cba

VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Nov 24 2016 16:43:18)
Included patches: 1-52
Extra patches: 8.0.0056

(Also seen on vim 8.2.318.)

Reproduced on minimal vimrc:

```
$ ls ~/.vim/bundle/
pathogen  sensible  snippets  ultisnips
$ cat ~/.vimrc
runtime bundle/pathogen/autoload/pathogen.vim
call pathogen#infect()

colorscheme desert
```

PR #1481, opened 2023-01-27
Fix Triple-quote snippet for python produces unbalanced string #1198

sharils:


lpil:
What does this do? Thanks

sharils:
Oops, @lpil , I forgot to mention that it fixes #1198

lpil (approved):
Ah! that makes sense. Than kyou

PR #1483, opened 2023-02-13
Support heex filetype

mxgrn:
The Elixir community recently added .heex files that are detected as a "heex" filetype. The snippets are mostly compatible with `eelixir` (legacy now), that's why this simple PR should make lives of many Elixir developers easier. Thanks!

### DIFF
--- a/snippets/heex.snippets
+++ b/snippets/heex.snippets
@@ -1,0 +1,1 @@
+extends eelixir


lpil (approved):
Nice!

By the way, `eelixir` isn't legacy. It's still part of Elixir core and will continue to be supported. It's fab in situations in which LiveView isn't suitable.

PR #1487, opened 2023-04-25
TeX: remove spurious line breaks

Aster89:
These line breaks in fact result in inserting a line break in the TeX source, which is definitely not desired for any of them.

It's pretty clear that those line breaks were added for "readability" at some change in the past, as the all start from line 393 and affect all subsequent snippets.


lpil (approved):
Thank you

PR #1491, opened 2023-08-27
elixir.snippets extending html to allow html completions in elixir files

krishna1m:
I was trying to write a HEEx template inside an elixir file inside the render function. I could not use the completions, doing this fixed it.

### DIFF
--- a/snippets/elixir.snippets
+++ b/snippets/elixir.snippets
@@ -1,3 +1,5 @@
+extends html
+
 snippet do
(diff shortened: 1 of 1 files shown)


lpil:
Hi! Thank you for the contribution but I don't think we should have this as a default as the majority of Elixir code does not embed HTML. Instead we can let the programmer decide if they want to opt-in to this behaviour

PR #1493, opened 2023-08-29
Support expect

Freed-Wu:


Freed-Wu:
expect is a tcl syntax. See vim runtime.

lpil (commented):

I couldn't find any documentation of this format online.

PR #1499, opened 2023-09-12
Scheme: improve alignment for some snippets and add one UltiSnips snippet

Aster89:


lpil (commented):
This will now be mixing tabs and spaces, is that desirable? Until this commit we've used tabs always as this means the user can configure indentation to whatever is their preference.

Aster89:
That's intendeed.

In scheme functions' and special forms' applications look syntactically the same, i.e. `(func-or-macro arg1 arg2 ...)`, so there's no reason to not align `arg1`, `arg2` and the rest under the same vertical.

Try to write the following, say, on [rextester](https://rextester.com/l/scheme_online_compiler) (or even on Vim itself; I haven't customised anything, so that should be the default), and press enter right before entering the marked character
```scm
(cond (#f "ciao") (#t "uffa"))
;     ^           ^

(cond (#f "ciao") (#t "uffa"))
;                 ^
```
and you'll get these:
```scm
(cond (#f "ciao")
      (#t "uffa"))

(cond
 (#f "ciao")
 (#t "uffa"))
```
The latter can all be handled by tabs, but the former can't, it must use spaces because the indentation must exactly be the same width as `(cond `.

I'm obviously assuming claiming that having the various "arguments" of functions/special forms be aligned vertically is the right thing to do.

lpil:
I understand the intention, what I mean is that with this if the programmer says they wish to indent using tabs we will no longer respect that wish. This is not what we do with any other snippets, including all the other Lisp favours.

Aster89:
@lpil , what if I try using `!v &expandtab` to decide whether to use tabs (so expanding to the original snippets) or spaces (thus using my proposed indentation)?

lpil:
Sorry, we use tabs in this project.

PR #1505, opened 2023-11-22
Tex: add modern textbf/textit/texttt commands

sanjayankur31:


sanjayankur31:
Question: is there value in making sure the ultisnip and snippets commands are the same. Or, on the contrary, is there a reason to make sure they're not the same?

lpil:
The ultisnips ones should be removed- they should only exist for specific snippets that cannot be implemented using the snipmate format.

lpil:
It is already documented, but improvements to the README are welcome.

sanjayankur31:
Updated to remove UltiSnip snippets that are already defined in the snipmate format.

I couldn't find where in the docs it says to prefer snipmat since UltiSnips can also read them, so I've added a line in the Readme.

UltiSnips/tex.snippets
```diff
@@ -467,112 +387,4 @@ snippet ac "Acroynm normal" bi
...
-\subparagraph{$1${VISUAL}}
-endsnippet
-
-snippet subpar* "Sub Paragraph*" wi
-\subparagraph*{$1${VISUAL}}
-endsnippet
-
-snippet chapl "Chapter with label" wi
-\chapter{$1${VISUAL}}
-\label{cha:${2:${1/\W+/-/g}}}
```
sanjayankur31, line 525: these are already defined in snipmate, so I removed them, but the labels here are more advanced, so we could leave these in as UltiSnips specific overrides.

lpil (approved):
Thank you

PR #1511, opened 2024-01-24
Update ruby.snippets: add binding.irb debug

r3jack:
["If you use Ruby 2.5 or later versions, you can also use binding.irb in your program as breakpoints."](https://github.com/ruby/irb?tab=readme-ov-file#the-bindingirb-breakpoint)


lpil (approved):
Thank you

Issue #1514, opened 2024-03-08
The commit that remove duplicate snippets broke important latex feature.


gzqx:
#1474 broke the feature introduced in #1505. Without the regrex substitution, it will put space inside label, which is a syntax error in latex.

According to the document of SnipMate it should support regrex from `|substitution|`, however I cannot get any snippet including the examples in the document to work. I am not sure if there is some special setting in `vim-snippet` which caused this or there is simply some bug in SnipMate. For now I would recommend bring back the snippets removed in #1474.

gzqx:
For some reason I don't know, I now managed to get it working with snipmate syntax. I have created a pull request #1515

gzqx:
However, notice that this syntax is not compatible with Ultisnip, which means that things will go wrong if one use ultisnip instead of snipmate. I still suggest bringing back #1474 in combination with this PR.

PR #1515, opened 2024-03-08
Add white space substitution for label in latex back to snipmate version

gzqx:


gzqx:
See #1514
However, notice that this syntax is not compatible with Ultisnip, which means that things will go wrong if one use ultisnip instead of snipmate. I still suggest bringing back https://github.com/honza/vim-snippets/pull/1474 in combination with this PR.

lpil:
Could you remove be change to the readme please

PR #1517, opened 2024-06-03
Fixed formatting of snippets file

ndrew222:


ndrew222:
Just realised the snippets folder is in SnipMate format, not UltiSnips

PR #1521, opened 2024-08-09
Remove the default protocol for markdown link creation

joshpetit:
I think it makes sense for the default behavior to be not adding a protocol to a link. 99% of the time when I'm adding a link into markdown I don't just copy the domain and path but the entire protocol. And sometimes my links are anchor tags or to other files. So removing the default `https:` protocol I believe makes the most sense.


lpil (approved):
Thank you

PR #1526, opened 2024-08-30
Support xf86conf filetype

Freed-Wu:


lpil (commented):

I couldn't find any documentation of this format online.

PR #1529, opened 2024-08-30
Support mysql filetype

Freed-Wu:


lpil (commented):

I think you may have committed the wrong file here?

PR #1541, opened 2025-09-03
tex: Remove unnumbered equation

fsimonis:
This PR removes the unnumbered equation snippet `snippet \`.

This snippet leads to a lot unwanted use, especially when using `\\` used very commonly for linebreaks in itemize.
I also interferes with verbatim envs where the target language uses `\` at the end of the line (Dockerfile, shell scripts, C/C++ Macros).

```tex
\begin{itemize}
  \item First line\\|
   second line
  \item etc
\end{itemize}
```


honza:
Thanks!

PR #1544, opened 2025-12-25
New JavaScript snippet: `for..of`

caiangums:
Similar to the `for..in` for objects, the `for..of` can be used with Iterable Objects such as `Array`, `Map`, and `Set`.
- [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

### DIFF
--- a/snippets/javascript/javascript.snippets
+++ b/snippets/javascript/javascript.snippets
@@ -123,6 +123,11 @@ snippet fori
 	for (let ${1:prop} in ${2:object}) {
 		${0:$2[$1]}
 	}
+# For of loop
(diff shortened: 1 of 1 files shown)


caiangums:
I found that this is already present at a different place in the `javascript.snippets` file ([this line](https://github.com/honza/vim-snippets/blob/ededcf7581962ee616cadab360d5966f3307f11a/snippets/javascript/javascript.snippets#L316)). I can change this to move this snippet to the `# Flow control` section of the file, or you can close this PR @honza 😄

